PR #10, opened 2022-03-21
Fix mobile menu on dark mode

nachoiacovino:


constants/currencies.ts
```diff
@@ -216,11 +216,6 @@ const Mumbai: CurrencyMetadata[] = [
     name: "Polygon",
     symbol: "MATIC",
   },
-  {
```
jnsdls, line 219: why are removing this again?

jnsdls (commented):
quick questions

PR #11, opened 2022-03-21
Update PostHog tracking

nachoiacovino:
Reasoning here: https://www.notion.so/thirdweb/Update-to-PostHog-analytics-for-v2-0a27c9aee4b04fa8abe4cbbc4464d5ca


jnsdls:
@nachoiacovino u created the necessary actions in PH to handle the switch?

jnsdls:
need to instead move to on-chain analytics

PR #26, opened 2022-03-28
Update CID on embeds

nachoiacovino:


jnsdls:
@nachoiacovino it looks like with MM I cannot claim in embeds / buy in marketplaces anymore after this update, please make sure that's an isolated issue for me

PR #32, opened 2022-03-29
Rename minter to creator

nachoiacovino:
Also update some descriptions

https://linear.app/thirdweb/issue/FTD-118/rename-minter-role-to-creator-role-on-drops

### DIFF
--- a/components/contract-tabs/permissions/ContractPermission.tsx
+++ b/components/contract-tabs/permissions/ContractPermission.tsx
@@ -47,7 +47,7 @@ export const ContractPermission: React.FC<IContractPermission> = ({
           <Flex>
             <Stack spacing={1} flexGrow={1}>
               <Heading size="heading.sm" textTransform="capitalize">
-                {role}
+                {role === "minter" ? "Creator" : role}
               </Heading>
               <Text>{description}</Text>
             </Stack>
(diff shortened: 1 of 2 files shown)


components/contract-tabs/permissions/ContractPermission.tsx
```diff
@@ -47,7 +47,7 @@ export const ContractPermission: React.FC<IContractPermission> = ({
           <Flex>
             <Stack spacing={1} flexGrow={1}>
               <Heading size="heading.sm" textTransform="capitalize">
-                {role}
```
jnsdls, line 50: this means we'd rename it for all contracts, is that intended? Or did we mean to only do it for drops?

jnsdls (commented):
Just 1 quick clarification

components/contract-tabs/permissions/ContractPermission.tsx
nachoiacovino: I think it's okay renaming for all contracts, if you believe it's better to do it only for drops, I can check contractType here and do it only then

nachoiacovino: I just thought about it again, and NFT Collection and Edition have signature based minting, which allows people to mint without the minter role, so it can also get confusing on those.

jnsdls: ok sounds good, if it becomes confusing we'll need to react but :shipit: for now

jnsdls (approved):
LGTM :shipit:

PR #38, opened 2022-03-31
initial setup for react 18

jnsdls:


nachoiacovino:
React 18 doesn't adds children to React.FC, so we would need to update every single component and add children: React.Node

Probably worth looking into using JSX.Element instead, more info here: https://github.com/facebook/create-react-app/pull/8177

jnsdls:
we'll wait for next to be stable with react 18 for this

PR #62, opened 2022-04-14
Update AVAX RPC 

nachoiacovino:


nachoiacovino:
will explore better options than chainstacks free tier, I have a call with chainstacks on Tuesday to talk about their paid solutions.

PR #63, opened 2022-04-14
Correct tokenId batch upload

nachoiacovino:
Fixes so it shows the correct token id when uploading a second batch


jnsdls (approved):
LGTM

PR #66, opened 2022-04-14
trait_type validation

nachoiacovino:


components/contract-pages/forms/mint/mint.edition.tsx
```diff
@@ -71,6 +72,13 @@ export const EditionMintForm: React.FC<IEditionMintForm> = ({ contract }) => {
       isClosable: true,
     });
   };
+
+  // TODO FIXME
+  const onSubmit = (data: any) => {
```
jnsdls, line 77: no reason to have this be an extra function (with the :any) just move this as a anon function directly into the handleSubmit() of the form

components/contract-pages/forms/mint/mint.nftcollection.tsx
```diff
@@ -79,7 +80,8 @@ export const NFTCollectionMintForm: React.FC<INFTCollectionMintForm> = ({
 
   // TODO FIXME
   const onSubmit = (data: any) => {
```
jnsdls, line 82: may as well fix this onSubmit function by just moving it into the form as well

utils/parseAttributes.ts
```diff
@@ -0,0 +1,15 @@
+export const parseAttributes = (
+  attributes: {
+    trait_type: string;
+    value: string;
+  }[],
+) => {
+  const parsedAttributes = attributes
+    .filter(({ value }) => value !== "")
```
jnsdls, line 8:
```suggestion
    .filter(({ value }) => value === 0 || value)
```

specifically check for `0` (which is valid) everything else (empty string, `null`, `undefined`) is invalid and should be filtered out

jnsdls (changes requested):

utils/parseAttributes.ts
nachoiacovino: with the single nft upload, 0 will actually be a string, so no need to check, example: https://thirdweb-r7lmo5xuh-thirdweb.vercel.app/dashboard/polygon/nft-drop/0x0dbC68329c91AAF7Bc849A27FF363DcBc38BBC55

components/contract-pages/forms/mint/mint.edition.tsx
nachoiacovino: the any is still necessary otherwise we get this
image

probably I'm missing something here

jnsdls (approved):

nachoiacovino:
Everything you did makes sense @jnsdls, but deleting the map made it so no values will return an empty array instead of undefined and only value will return empty trait_type, which we don't want
image

Added back the .map to get the desired result:
image

PR #75, opened 2022-04-22
Add maxClaimable to snapshots

nachoiacovino:


components/batch-upload/SnapshotUpload.tsx
```diff
@@ -289,7 +300,13 @@ const SnapshotTable: React.FC<SnapshotTableProps> = ({ data, portalRef }) => {
           }
         },
       },
-    ] as Column<{ address: string }>[];
+      {
+        Header: "Max claimable per transaction",
+        accessor: ({ maxClaimable }) => {
+          return maxClaimable || "Default";
```
jnsdls, line 306: probably just ` || 1` here, no? (that *is* the default?)

jnsdls (changes requested):

components/batch-upload/SnapshotUpload.tsx
nachoiacovino: No it's not, it defaults to whatever you have set on max claimable in the claim phase
image

jnsdls: ah, fair enough then

jnsdls (approved):

PR #79, opened 2022-04-23
Add ERC-1155 airdrop

nachoiacovino:


jnsdls (approved):
shippable, some quick notes:

- we should still *detect* if the total number of things the user wants to airdrop is higher than what they own (and disable the airdrop button in that case)
- we now have a lot of the "same" bulk upload tables that function very similarly (just have different columns maybe) and we should consider unifying them

_but neither of those are blockers_

PR #82, opened 2022-04-25
Add map_image to batch upload

nachoiacovino:
@saminacodes please test it out and tell me if this is what you had in mind.

What I really like about this is for people that want to use the same image for multiple NFTs, they can do so very easily like this.


saminacodes:
CleanShot 2022-04-25 at 09 18 38@2x
map_image works well - would add the property in the csv file example so people don't miss it. Checking map_animation_url here in a second

jnsdls:
why are we calling it `map_` instead of just using the normal keys?

jnsdls (approved):
LGTM

PR #85, opened 2022-04-25
Add burn action to NFTs

nachoiacovino:


@3rdweb-sdk/react/hooks/useCommon.ts
```diff
@@ -375,3 +380,56 @@ export function useAirdropMutation<TContract extends ValidContractInstance>(
...
+                  return cacheKeys.list(contract?.getAddress());
+                }
+                return undefined as never;
+              })
+              .filter((fn) => !!fn),
+          );
+        }
+
+        return invalidate([
+          CacheKeyMap[contractType].list(contract?.getAddress()),
```
jnsdls, line 430: this needs to also invalidate the `supply` query ("supply" goes down when burning)

@3rdweb-sdk/react/hooks/useCommon.ts
```diff
@@ -375,3 +380,56 @@ export function useAirdropMutation<TContract extends ValidContractInstance>(
...
+      onSuccess: (_data, _variables, _options, invalidate) => {
+        // this should not be possible, but we need to catch it in case it does
+        // if we don't know we just invalidate everything.
+        if (!contractType) {
+          return invalidate(
+            Object.keys(CacheKeyMap)
+              .map((key) => {
+                const cacheKeys = CacheKeyMap[key as keyof typeof CacheKeyMap];
+                if ("list" in cacheKeys) {
+                  return cacheKeys.list(contract?.getAddress());
```
jnsdls, line 421: same here, needs to invalidate supply query on all contracts in the fail case probably

jnsdls (changes requested):
almost there, just fix invalidation

@3rdweb-sdk/react/hooks/useCommon.ts
```diff
@@ -375,3 +380,59 @@ export function useAirdropMutation<TContract extends ValidContractInstance>(
...
+                  return cacheKeys.supply(contract?.getAddress());
+                }
+                return undefined as never;
+              })
+              .filter((fn) => !!fn),
+          );
+        }
+
+        return invalidate([
+          CacheKeyMap[contractType].list(contract?.getAddress()),
```
jnsdls, line 433: what about supply here?

jnsdls (changes requested):
need to check if supply exists and invalidate it

@3rdweb-sdk/react/hooks/useCommon.ts
nachoiacovino: supply only exists in NFTDrop contracts, if I add it here it gives an error on all the other contracts, trying to make sense on how to refresh on Edition in which it shows

jnsdls: we definitely *have* supply numbers for them, so find the useQuery for that they're using and make sure they are being invalidated

jnsdls (approved):

PR #96, opened 2022-04-28
Add airdrop tracking

nachoiacovino:


components/contract-pages/table/expansions/AirdropSection.tsx
```diff
@@ -41,30 +45,35 @@ export const AirdropSection: React.FC<IAirdropSection> = ({
...
+              onError,
+              onSuccess: () => {
+                onSuccess();
+                trackEvent({
+                  category: "airdrop",
+                  action: "success",
+                  label: "edition",
+                  contractAddress: contract?.getAddress(),
+                  tokenId,
+                  path: asPath,
```
jnsdls, line 69: couple of things here:
- `category` should be the most top level part of the structure, you will end up looking this up by looking at `category.action.label`
- `action` should be the user triggered type of event, so in this case `"airdrop"` would make sense here
- `label` should be a thing that you can then filter by (so in this case `success` would make sense here)

- contractAddress and tokenId make sense, path is not useful because every event will already include the full url etc

```suggestion
                  category: "edition",
                  action: "airdrop",
                  label: "success",
                  contractAddress: contract?.getAddress(),
                  tokenId
```

jnsdls (changes requested):
small changes

jnsdls (approved):

PR #120, opened 2022-05-15
Fix Coinbase Wallet not showing on mobile

nachoiacovino:
This fixes the issue but makes it so others like Trust wallet show like this: 

image

Seems like wagmi returns "Injected" for trust specifically, since on 0.2.x it was not on their list:

https://github.com/tmm/wagmi/blob/0.2.x/packages/core/src/utils/getInjectedName.ts

this is fixed on wagmi 0.3.x:

https://github.com/tmm/wagmi/blob/main/packages/core/src/utils/getInjectedName.ts

We can merge this then improve it and do the whole map to add an icon once we update to wagmi 0.3.x

One quick fix would be add an icon for when it's "Injected", so it's not blank.


jnsdls:
@nachoiacovino  yeah as you mention the real fix for this will come with the updated wagmi (via upgraded react) package

for now let's do 2 things:
1: if id is `injected` show a prettier name than "Injected"  - maybe "Browser Extension"
2: add a generic icon for that case

then let's ship it

PR #128, opened 2022-05-18
Update privacy/tos notice

nachoiacovino:


jnsdls:
we'll do this differently

PR #158, opened 2022-05-31
Add balance zero notice

nachoiacovino:
We need the `useBalance` hook to invalidate on network switch to be able to merge this.


jnsdls (approved):
just add some basic tracking, then ship

PR #181, opened 2022-06-08
Fix buggy prompt on claim phases

nachoiacovino:


nachoiacovino:
merging for now, not perfect but better and we'll finish the fix later

PR #211, opened 2022-06-27
activity feed

jnsdls:


jnsdls:
> * height should wrap / be shorter  (contract interaction widget should be on top, gets pushed really low)
> * popover has transparent background?
> 
> Screen Shot 2022-06-27 at 2 55 21 PM
> 
> * overall unsure about the popover, feels a bit clunky to explore my events. Maybe accordeon that keeps stuff open would feel better?
> 
> btw i'm down to help on the UI here if you want, just polishing it a bit (the interaction widget too)

I fixed the obvious issues just now, will ship but feel free to jump in and clean things up as you find them anytime

PR #212, opened 2022-06-27
Add audit links

nachoiacovino:


joaquim-verges:
@nachoiacovino where do the "audited" labels show up ? can't find them

jnsdls:
the entire flow can improve further, we should kick off another round of design on it. that said this is much better than not having the audit information so I'll ship this now.

PR #215, opened 2022-06-28
Enumerable check

nachoiacovino:
Not sure about the implementation here.

Right now we're checking ERC721/ERC1155 for the tab and giving the user a checkmark, but we actually care about Enumerable for that tab to show anything.

might be worth showing the tab always but without the checkmark? or tab always but checkmark when enumerable?

for now, I left it so it shows the tab with checkmark when ERC721/ERC1155 but it tells you that you need to implement enumerable, also hides the Mint button if the contract is not mintable


jnsdls:
implementation makes sense to me, we might be able to clean up the underlying logic of "is a feature enabled" more but that's not part of this arc of work

jnsdls (approved):
implementation makes sense to me, we might be able to clean up the underlying logic of "is a feature enabled" more but that's not part of this arc of work

PR #220, opened 2022-06-30
Parse array and tuple inputs on contract code component

adam-maj:


contract-ui/tabs/overview/cards/custom-contract-code.tsx
```diff
@@ -228,9 +229,7 @@ const InteractiveAbiFunction: React.FC<InteractiveAbiFunctionProps> = ({
     data,
     error,
     isLoading: mutationLoading,
-  } = useMutation(async (params: unknown[] = []) =>
-    abiFunction ? await contract?.call(abiFunction.name, ...params) : undefined,
-  );
+  } = useMutation(async (params: unknown[] = []) => contractCall(params));
```
jnsdls, line 232: can we move this below the definition of `contractCall()` (I know it may work without doing that but it hurts my brain to use a function above where we declare it)

contract-ui/tabs/overview/cards/custom-contract-code.tsx
```diff
@@ -261,6 +260,66 @@ const InteractiveAbiFunction: React.FC<InteractiveAbiFunctionProps> = ({
...
+        if (Array.isArray(parsed) || typeof parsed === "object") {
+          return parsed;
+        } else {
+          // Return original value if its not an array or object
+          return p;
+        }
+      } catch {
+        // JSON.parse on string will throw an error
+        return p;
+      }
```
jnsdls, line 280: _I wish we could do this somehow cleaner, but not a blocker_

contract-ui/tabs/overview/cards/custom-contract-code.tsx
```diff
@@ -261,6 +260,66 @@ const InteractiveAbiFunction: React.FC<InteractiveAbiFunctionProps> = ({
...
+      const parsed = JSON.parse(param.value);
+      if (Array.isArray(parsed) || typeof parsed === "object") {
+        return param.value;
+      } else {
+        return displayString(param.value);
+      }
+    } catch {
+      return displayString(param.value);
+    }
+  }
```
jnsdls, line 321: since this functions work directly on their inputs and don't have side-effects can we move them out of the component?

these are basically util functions, they don't need to exist inside the component itself

jnsdls (changes requested):
couple of small changes please. should be quick! :)

joaquim-verges (approved):
fixes the issue so let's ship it. gonna have to get some design help to make these inputs real nice and flexible in the future, but at least now ppl can actually execute their functions

jnsdls (approved):

PR #224, opened 2022-07-01
Fix multiwrap getAll

nachoiacovino:


jnsdls (approved):
LGTM, as a follow-up lets update the typescript & react SDK to make pagination for multi-wrap work

PR #227, opened 2022-07-04
Update navigation links

adam-maj:


nachoiacovino:
Pricing is the most clicked thing after Start building, not sure we should kill it, dashboard would send people to the same place as Start building, what's the motivation for adding it?

And in case we want exactly this behavior, dashboard link should be relative not absolute

adam-maj:
@nachoiacovino Is there a reason we want that to be focusing on our pricing though? People may be clicking it for a number of reasons (ie. if it costs money they can't use it, make sure it's affordable, etc. - but number of clicks reflects peoples curiosity, not necessarily importance for us to highlight). If anything people can forget about pricing and just focus on what they can do with our stuff

nachoiacovino (changes requested):
- Menu only closes when clicking out (this might be okay)
- You can only click on the text instead of the whole MenuItem (blocker)

joaquim-verges:
@adam-maj let's adress the feedback and take this to the finish line. I think it's a good change. Maybe check with @furqanrydhan or @jakeloo for the pricing top link removal tho

PR #242, opened 2022-07-07
Add option to randomize batch upload

nachoiacovino:


eabdelmoneim:
Can we make the check box for randomization be at the top of the Delayed Reveal panel instead of at the bottom?

I don't like the text Randomize the order of the NFTs. (Off-chain randomization)

PR #259, opened 2022-07-13
Auto generating thumbnail

nachoiacovino:


joaquim-verges:
how do i test this? tried sharing via twitter but no thumbnail showed up

nachoiacovino:
opengraph.xyz works for me

jnsdls:
superseded by #269 (but we'll still use the thumbnail generator)

PR #271, opened 2022-07-19
Fix IPFS upload bug

adam-maj:


jnsdls (approved):
tested deploying contracts, batch upload & single create of NFTs. all worked for me

PR #278, opened 2022-07-19
enable sig drop on mainnet

jnsdls:


nachoiacovino (approved):
tested with polygon, all good!

PR #283, opened 2022-07-20
Add tracking to release pages

nachoiacovino:


jnsdls (approved):
this makes sense to me so far, we need to align on what the expected user-flows are and how these interactions will show up in them, but it's a good start

nachoiacovino:
Superseded by #338

PR #284, opened 2022-07-20
Add avatar edit

nachoiacovino:


components/contract-components/releaser/releaser-header.tsx
```diff
@@ -34,12 +36,23 @@ export const ReleaserHeader: React.FC<ReleaserHeaderProps> = ({
...
+              src={releaserProfile.data?.avatar}
+            />
+          ) : (
+            <ChakraNextImage
+              alt=""
+              boxSize={12}
+              width="100%"
+              height="100%"
+              mt={1}
+              src={`https://source.boringavatars.com/marble/120/${wallet}?colors=264653,2a9d8f,e9c46a,f4a261,e76f51`}
```
jnsdls, line 53: I think we can just use a regular old image for this, instead. Meaning we don't have to optimize these images.

next.config.js
```diff
@@ -80,6 +80,7 @@ const moduleExports = {
       "ipfs.thirdweb.com",
       "portal.thirdweb.com",
       "blog.thirdweb.com",
+      "source.boringavatars.com",
```
jnsdls, line 83: I don't think we need to add this if we just use a regular image

jnsdls (commented):
couple of comments

jnsdls (approved):
LGTM

PR #289, opened 2022-07-21
Markdown prettify

nachoiacovino:


components/contract-components/released-contract/markdown-renderer.tsx
```diff
@@ -11,29 +11,76 @@ import {
...
+        h5: (props) => (
+          <Heading size="subtitle.sm" {...commonHeadingProps} {...props} />
+        ),
+        h6: (props) => (
+          <Heading size="label.md" {...commonHeadingProps} {...props} />
+        ),
+        pre: (props) => (
+          <CodeBlock
+            code={onlyText(props.children).trim()}
+            language="jsx"
```
jnsdls, line 69: nah, we need to get this from the props if we can, can definitely not be generic `JSX`

jnsdls: this might be available from props if it's ```jsx style ? have to dig into it and see if it is available.

components/contract-components/contract-release-form/index.tsx
```diff
@@ -75,6 +90,15 @@ export const ContractReleaseForm: React.FC<ContractReleaseFormProps> = ({
 
   const ensName = useEnsName(address);
 
+  const placeholderVersion = useMemo(() => {
+    if (latestVersion) {
+      return `${latestVersion.split(".")[0]}.${
+        Number(latestVersion.split(".")[1]) + 1
+      }.${latestVersion.split(".")[2]}`;
+    }
```
jnsdls, line 98: wait so we bump the minor version by default? this seems odd, no?

jnsdls (changes requested):
couple of changes to look into please

PR #293, opened 2022-07-22
Make CSV use OS standard

nachoiacovino:
CSV upload wasn't using OpenSea standard, and would break OpenSea if the attribute column was empty, this standarizes to use OS standard everywhere and fixes that bug


utils/batch.ts
```diff
@@ -95,6 +96,21 @@ export const removeEmptyKeysFromObject = (obj: any) => {
...
+  const keys = Object.keys(obj || {});
+  const values = Object.values(obj || {});
+  const attributes = [];
+
+  for (let i = 0; i < keys.length; i++) {
+    attributes.push({
+      trait_type: keys[i],
+      value: values[i],
+    });
+  }
```
jnsdls, line 109: could be replaced with this

```suggestion
  const attributes = Object.entries(obj || {}).map(([trait_type, value]) => ({
    trait_type,
    value,
  }));
```

utils/parseAttributes.ts
```diff
@@ -11,16 +11,16 @@ export function parseAttributes<
...
   if (!data) {
     return data;
   }
   if (Array.isArray(data)) {
     const parsedArray = data
       .filter(({ value }) => value !== "")
       .map(({ value, trait_type }) => ({
-        value,
         ...(!!trait_type && { trait_type }),
+        value,
```
jnsdls, line 23: why change the order here?

jnsdls (commented):
some quick questions / suggestions

utils/parseAttributes.ts
nachoiacovino: makes more sense in the preview of the batch upload seeing trait_type first

utils/batch.ts
nachoiacovino: well that's miles easier, thanks

utils/parseAttributes.ts
jnsdls: ah so just visual ordering, ok

jnsdls (approved):
LGTM

PR #296, opened 2022-07-25
Send chain_id

nachoiacovino:


jnsdls (approved):
confirm with @thleo, but I think this is what we want

PR #306, opened 2022-07-26
Fix code suggestion on custom contracts

nachoiacovino:
This fixes a wrong suggestion of chainName on custom contracts

image

We should also merge this one: https://github.com/thirdweb-dev/typescript-sdk/pull/527 on the ts sdk that's not using the same mapping as in our dashboard


jnsdls (approved):
LGTM

PR #312, opened 2022-07-28
Fix code tab on custom contracts

nachoiacovino:


contract-ui/tabs/code/components/code-overview.tsx
```diff
@@ -12,88 +14,89 @@ interface CodeOverviewProps {
 const COMMANDS = {
   install: {
     javascript: "npm install @thirdweb-dev/sdk",
-    react: "npm install @thirdweb-dev/react",
+    react: "npm install @thirdweb-dev/react @thirdweb-dev/sdk ethers",
```
jnsdls, line 17: so we add more here but we don't add `ethers` to the javascript install at the same time?

jnsdls (commented):
hm

jnsdls (approved):

PR #313, opened 2022-07-28
Trim quantity/maxClaimable

nachoiacovino:
Also, add clarification that repeated addresses will get removed

### DIFF
--- a/components/batch-upload/AirdropUpload.tsx
+++ b/components/batch-upload/AirdropUpload.tsx
@@ -122,8 +122,6 @@ export const AirdropUpload: React.FC<AirdropUploadProps> = ({
     return ordered;
   }, [validAirdrop]);
 
-  console.log(data);
-
   const { getRootProps, getInputProps, isDragActive } = useDropzone({
     onDrop,
   });
@@ -200,6 +198,10 @@ export const AirdropUpload: React.FC<AirdropUploadProps> = ({
                       <Link download color="blue.500" href="/airdrop.csv">
                         Download an example CSV
                       </Link>
+                    </ListItem>
+                    <ListItem>
+                      Repeated addresses will be deleted and only the first
+                      found will be kept.
                     </ListItem>
                   </UnorderedList>
                 </Flex>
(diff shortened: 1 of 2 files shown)


components/batch-upload/SnapshotUpload.tsx
```diff
@@ -90,7 +90,10 @@ export const SnapshotUpload: React.FC<SnapshotUploadProps> = ({
           const data: SnapshotAddressInput[] = (
             results.data as SnapshotAddressInput[]
           )
-            .map((a) => ({ ...a, address: (a.address || "").trim() }))
+            .map(({ address, maxClaimable }) => ({
+              address: (address || "").trim(),
+              quantity: maxClaimable ? (maxClaimable || "").trim() : "0",
```
jnsdls, line 95: are we sure we want to set "0" as the default instead of just not having the key?

components/batch-upload/SnapshotUpload.tsx
```diff
@@ -213,6 +216,10 @@ export const SnapshotUpload: React.FC<SnapshotUploadProps> = ({
                         Download an example CSV with maxClaimable
                       </Link>
                     </ListItem>
+                    <ListItem>
+                      Repeated addresses will be deleted and only the first
+                      found will be kept.
```
jnsdls, line 221: "deleted" -> "removed" (deleted sounds agressive)

components/batch-upload/AirdropUpload.tsx
```diff
@@ -196,6 +199,10 @@ export const AirdropUpload: React.FC<AirdropUploadProps> = ({
                         Download an example CSV
                       </Link>
                     </ListItem>
+                    <ListItem>
+                      Repeated addresses will be deleted and only the first
+                      found will be kept.
```
jnsdls, line 204: "deleted" -> "removed" (deleted sounds agressive)

jnsdls (changes requested):

components/batch-upload/SnapshotUpload.tsx
nachoiacovino: yes, "0" is what defaults to the one set in the claim phases, it's already the default

jnsdls (approved):

PR #317, opened 2022-08-01
Deploy drawer

nachoiacovino:
I also made the "Deploy now" to be in the same line as the select, fixing the weird spacing


jnsdls (changes requested):
image

I would expect this to conform the the grid layout (aka the select should not overflow the "released by" column imo


menu dropdown is overflowing the right side of the screen, can we fix that?

image

jnsdls (approved):
LGTM

PR #336, opened 2022-08-04
NFT drawer

nachoiacovino:
Transfer working for both erc721 and erc1155, good base for when we want to add more actions.

Also Airdrop working for ERC1155.


jnsdls (approved):
LGTM

PR #354, opened 2022-08-10
Add events tab to contract panel

nachoiacovino:


components/contract-functions/contract-function.tsx
```diff
@@ -110,30 +115,35 @@ export const ContractFunction: React.FC<ContractFunctionProps> = ({
   );
 };
 
-interface ContractFunctionsPanelProps {
-  functions: AbiFunction[];
-  contract?: SmartContract;
-}
+type ContractFunctionsPanelProps =
+  | { functions: AbiFunction[]; events?: undefined; contract?: SmartContract }
+  | { functions?: undefined; events: AbiEvent[]; contract?: SmartContract };
```
jnsdls, line 120: why can't we just use the same input? you're merging them below anyways

you can rename it from "functions" to "input" or "fnsOrEvents" or anythigne lse really

components/contract-functions/contract-function.tsx
```diff
@@ -149,25 +159,25 @@ export const ContractFunctionsPanel: React.FC<ContractFunctionsPanelProps> = ({
...
-                      fn.stateMutability === "view" ||
-                      fn.stateMutability === "pure"
-                        ? FiEye
-                        : FiEdit2
+                      isFunction
+                        ? (fn as AbiFunction).stateMutability === "view" ||
+                          (fn as AbiFunction).stateMutability === "pure"
+                          ? FiEye
+                          : FiEdit2
+                        : BsLightningCharge
```
jnsdls, line 176: this is not gonna work, this will mean it's always an `FiEye` for events except for the selected event?

components/contract-functions/contract-function.tsx
```diff
@@ -149,25 +159,25 @@ export const ContractFunctionsPanel: React.FC<ContractFunctionsPanelProps> = ({
           pr={{ base: 0, md: 3 }}
           mb={{ base: 3, md: 0 }}
         >
-          {sortedFunctions.map((fn) => (
-            <ListItem key={fn.signature} my={0.5}>
+          {fnsOrEvents.map((fn) => (
+            <ListItem key={fn.name} my={0.5}>
```
jnsdls, line 163: `name` is not good enough as a key, there are functions that may have the same name which will then clash

components/contract-functions/contract-function.tsx
```diff
@@ -33,13 +34,17 @@ export const ContractFunction: React.FC<ContractFunctionProps> = ({
     return null;
   }
 
+  const isFunction = "stateMutability" in fn;
```
jnsdls, line 37: I think I would prefer if we directly pass through if were in "function" or "event" mode

jnsdls (changes requested):
couple of things to look at

components/contract-functions/contract-function.tsx
nachoiacovino: I tested it and it works... not sure what you mean

components/contract-functions/contract-function.tsx
nachoiacovino: right but events don't have signature, just makes it annoying then to use the key

components/contract-functions/contract-function.tsx
jnsdls: oh, the giant ternary statement confused me. hmmm

components/contract-functions/contract-function.tsx
jnsdls: can u do: `key={fn?.signature || fn.name}` ? would solve that problem, no?

joaquim-verges:
Visually this looks good to me - check out my nice commented Events :)

https://thirdweb-www-git-ni-add-events-tab-thirdweb.vercel.app/joenrv.eth/EthroneContract

@nachoiacovino let's get @jnsdls happy and ship this pls

components/contract-functions/contract-function.tsx
jnsdls: @nachoiacovino this part is still a blocker

nachoiacovino: can't do that but I'll do the check

PR #359, opened 2022-08-10
Add lazymint and batch upload to ERC1155 contracts

nachoiacovino:


jnsdls (approved):
let's fix the LGTM then -> LGTM

PR #366, opened 2022-08-11
Prebuilt source tab

nachoiacovino:
I ended up fetching the source code from GitHub for each of the prebuilts, I like how it turned out, makes it pretty seamless.


constants/mappings.ts
```diff
@@ -89,6 +90,8 @@ export const BuiltinContractMap: Record<ContractType, BuiltinContractDetails> =
       audit: "QmNgNaLwzgMxcx9r6qDvJmTFam6xxUxX7Vp8E99oRt7i74",
       href: "/contracts/new/pre-built/drop/nft-drop",
       roles: ["admin", "transfer", "minter"],
+      sourceUrl:
+        "https://raw.githubusercontent.com/thirdweb-dev/contracts/main/contracts/drop/DropERC721.sol",
```
jnsdls, line 94: let's not pull from the `main` branch but instead from a specific commit hash (the one that is currently deployed and active)

@nachoiacovino please ask @nkrishang to provide that commit hash for each contract and then use that for each of them.

jnsdls (changes requested):
Can't use `main` branch for this, other than that a good approach

jnsdls (approved):
LGTM

PR #387, opened 2022-08-15
Add reset flag to make signature drop single phase consistent

nachoiacovino:


jnsdls (commented):
@nachoiacovino can u give context for this / how to test?

not sure I understand what is going on in this PR :)

nachoiacovino:
> @nachoiacovino can u give context for this / how to test?
> 
> not sure I understand what is going on in this PR :)

Right now, you cannot save an empty claim phase on Signature Drop, you need to reset claim eligibility and then remove, this fixes that.

Also, this fixes a problem in which Signature Drop behavior is not consistent with NFT Drop.

Example on NFT Drop: 
- I set it so 200 can be claimed.
- 100 are claimed.
- I click the bin button, and I set it so 1000 can be claimed.
- Now 1000 can be claimed.

These same steps on Signature Drop would make it after setting to 1000, only 900 can be claimed, since it's _the same phase_, but this PR makes it so clicking the bin (removing and "setting" a new one) would make it so it actually resets and you can claim 1000 (like people would expect)

PR #389, opened 2022-08-16
Add isAdmin check on settings for prebuilts

nachoiacovino:


contract-ui/tabs/settings/components/metadata.tsx
```diff
@@ -154,22 +160,22 @@ export const SettingsMetadata = <TContract extends PotentialContractInstance>({
           </Flex>
         </Flex>
 
-        {/*         <AdminOnly contract={contract}> */}
```
jnsdls, line 157: question:
doesn't `AdminOnly` just make the same `useIsAdmin()` check underneath? if so why do we need to change this?

jnsdls: just checked and it doesn't, but we could *make* it use `useIsAdmin()` underneath and then we can change it in a better place?

jnsdls (changes requested):
let's:
- change how the `<AdminOnly>` component behaves
- then just use the `<AdminOnly>` component like we had before

jnsdls (approved):
LGTM

PR #395, opened 2022-08-18
appURI extension

furqanrydhan:


contract-ui/tabs/settings/page.tsx
```diff
@@ -34,6 +34,13 @@ export const CustomContractSettingsTab: React.FC<
     feature: "PlatformFee",
   });
 
+  const detectedAppURI = extensionDetectedState({
+    contract,
+    feature: "AppURI",
+  });
```
vishal2612200, line 40: Hi Guy's, 

There is a lot of reductant code in this repo, you can reduce the number of repetitive code, If you could add ```currying``` to ```extensionDetectedState ```  and then pass ```useContract(contractAddress)``` as a single argument to pipe function with few modifications, you can use below as a pipe sample function.

```const pipe = (...func) => (value) => func.reduce((acc, fn) => [...acc, fn(value)], [])```


Then use the output of the pipe function for your if statement and for react, component part also to remove repetitive code.
It will further reduce your reductant code.

File: contract-ui/tabs/settings/page.tsx [hunk 2]
```diff
@@ -43,7 +50,8 @@ export const CustomContractSettingsTab: React.FC<
     detectedMetadata === "disabled" &&
     detectedPrimarySale === "disabled" &&
     detectedPlatformFees === "disabled" &&
-    detectedRoyalties === "disabled"
+    detectedRoyalties === "disabled" &&
+    detectedAppURI === "disabled"
```
vishal2612200, line 54: repetitive code, store ```disabled``` as a constant value at a single place.
You can remove all these extra variables with the pipe function output, you can utilize the power of ```some``` here.

A lot of repetitive code in react components also. Can't comment on them as you can comment on current changes only.

PR #406, opened 2022-08-23
Update how we get contract explorer

nachoiacovino:


jnsdls:
I like this, we should remove the functions getter logic from the react SDK in general, it's an internal dashboard need not a react sdk need (was my bad to add it there in the first place)

jnsdls (approved):
LGTM

PR #424, opened 2022-08-25
always show verify on etherscan button

joaquim-verges:
if already verified, it'll tell you once you click the button
(we want this to verify our own implementations we deploy)

as a next step we can fetch verification status first and show a "already verified" label instead


jnsdls (approved):
LGTM

PR #429, opened 2022-08-26
Fix infinite skeleton on first release

nachoiacovino:


components/contract-components/contract-table/cells/released-by.tsx
```diff
@@ -26,8 +26,17 @@ export const ContractReleasedByCell: React.FC<
       .publisher;
 
   const releaserEnsQuery = ens.useQuery(releaser);
+
+  console.log({ isSuccess: fullPublishMetadata, isPrebuilt });
```
jnsdls, line 30: let's remove this

components/contract-components/contract-table/cells/released-by.tsx
```diff
@@ -26,8 +26,17 @@ export const ContractReleasedByCell: React.FC<
...
+
   return (
-    <Skeleton isLoaded={fullPublishMetadata.isSuccess || isPrebuilt}>
+    <Skeleton
+      isLoaded={
+        fullPublishMetadata.isSuccess ||
+        fullPublishMetadata.isStale ||
```
jnsdls, line 36: why does the `isStale` get hit?

components/contract-components/contract-table/cells/description.tsx
```diff
@@ -23,12 +23,18 @@ export const ContractDescriptionCell: React.FC<
     BuiltinContractMap[value as keyof typeof BuiltinContractMap]?.description;
 
   return (
-    <Skeleton isLoaded={fullPublishMetadata.isSuccess || !!description}>
+    <Skeleton
+      isLoaded={
+        fullPublishMetadata.isSuccess ||
+        fullPublishMetadata.isStale ||
```
jnsdls, line 29: why does the isStale get hit, how does this fix it?

jnsdls (changes requested):
remove the log and explain why `isStale` fixes it pls

components/contract-components/contract-table/cells/released-by.tsx
nachoiacovino: so we never get fullPublishMetadata if the contract has never been released before, is there a better way to check this?

jnsdls: ah, as in we never try to query it or we try to query but it returns a 404?

stale => there is data in the cache but it needs to be re-fetched (which seems like the wrong thing to check)

we could check isIdle *maybe*? or if we try to query and it's a 404 we can check the error?

nachoiacovino: I just updated to check that we're not fetching, this works better I can try again with isIdle see if that works better, I'll brb in 30 mins

jnsdls (approved):
this should work now, @nachoiacovino take a look if isIdle might be better still but will get this in for now

PR #479, opened 2022-09-13
Show more on readme/changelog

nachoiacovino:
Right now works fine for long readme/changelog but still shows the button on short ones


jnsdls:
we'll get back to it later if it comes back up

PR #494, opened 2022-09-22
Add storage dev build

adam-maj:


adam-maj:
confirmed - progress bar working with this

PR #501, opened 2022-09-26
[WIP] Add storage product page

adam-maj:


jnsdls:
@adam-maj while ur at it can u update the images for the other product pages (where available) for this as well

PR #510, opened 2022-09-28
Disable deprecated networks deployment

nachoiacovino:
And show deprecation notices.


components/notices/ChainDeprecation.tsx
```diff
@@ -0,0 +1,32 @@
+import { useActiveChainId } from "@3rdweb-sdk/react";
+import { Flex } from "@chakra-ui/react";
+import { SUPPORTED_CHAIN_ID } from "@thirdweb-dev/sdk";
+import { deprecatedChains } from "constants/mappings";
+import { Text } from "tw-components";
+
+export const ChainDeprecation = () => {
```
jnsdls, line 7: Should be React.FC but this will work

components/notices/ChainDeprecation.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const currentDate = new Date();
+  const deprecationDate = new Date("2021-10-05T00:00:00.000Z");
```
jnsdls, line 9: This will work but I would typically set it up outside of the component so it doesn't get re-instantiated on every render

components/selects/SupportedNetworkSelect.tsx
```diff
@@ -63,8 +62,10 @@ export const SupportedNetworkSelect = forwardRef<
               disabled={disabledChainIds?.includes(tn.chainId)}
             >
               {tn.chainName} ({tn.symbol})
-              {deprecatedChains.includes(tn.chainId) && " - Deprecated"}
-              {disabledChainIds?.includes(tn.chainId)
+              {deprecatedChains.includes(tn.chainId as SUPPORTED_CHAIN_ID) &&
```
jnsdls, line 65: I'd prefer to move this to a separate deprecated option group below, then we remove it fully after the deprecation date.

constants/mappings.ts
```diff
@@ -40,21 +41,28 @@ export interface BuiltinContractDetails {
...
-  token: [],
-  vote: [],
-  marketplace: [],
-  pack: [],
-  split: [],
-  "token-drop": [],
-  "signature-drop": [],
-  multiwrap: [],
-  custom: [],
+  "nft-drop": [...deprecatedChains],
```
jnsdls, line 51: Nicely done

jnsdls (approved):
Couple of things that could be cleaned up but overall good to go.

PR #516, opened 2022-09-29
Check if image/animation_url is mapped to not assign by order

nachoiacovino:


eabdelmoneim:
tested with the json I had issues with and it is working

nachoiacovino:
> tested with the json I had issues with and it is working

awesome, shipping

PR #486, opened 2022-09-14
init solana dashboard

jnsdls:

PR #518, opened 2022-09-29
feat(faucet): build solana faucet

avneesh0612:
## Features built

- basic UI for the faucet
- backend for requesting airdrop of SOL tokens

## Features to be built

- Add FAQ section

image


components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
...
+            _hover={{
+              bg: "rgba(255, 255, 255, 0.8)",
+            }}
+            bg="white"
+            textColor="black"
+            px="6"
+          >
+            Start Building
+          </Button>
+        </Link>
```
jnsdls, line 35: we have a `<LinkButton>` component for this behavior, you can probably copy the way it works from one of the product pages (`/pages/auth.tsx` for example)

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Button, Text } from "tw-components";
+
+const CTA: FC = () => {
+  return (
+    <Flex
+      bg="url(/assets/faucet/cta-bg.png) no-repeat center"
+      bgSize="cover"
+      flexDir="column"
+      gap="6"
+      h="226px"
```
jnsdls, line 13: not loving the hard coded height here, can we use an `<AspectRatio>` to achieve the same goal? this hard-coded height means this banner will be huge on mobile

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
...
+      justify="center"
+      mt="10"
+      px="10"
```
jnsdls, line 16: preference only: we usually have these as numbers

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
+import { Flex } from "@chakra-ui/react";
+import Link from "next/link";
+import type { FC } from "react";
+import { Button, Text } from "tw-components";
+
+const CTA: FC = () => {
+  return (
+    <Flex
+      bg="url(/assets/faucet/cta-bg.png) no-repeat center"
+      bgSize="cover"
```
jnsdls, line 10: I think you can achieve this entire piece better by using our `<Card>` component (and then just giving it the bg image)

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
...
+        </Link>
+        <Link href="https://portal.thirdweb.com" passHref>
+          <Button
+            bg="rgba(255, 255, 255, 0.1)"
+            border="1px solid rgba(255, 255, 255, 0.2)"
+            px="6"
+          >
+            View docs
+          </Button>
+        </Link>
```
jnsdls, line 44: should be a `<LinkButton>` again :)

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </Flex>
+    </Flex>
+  );
+};
+export default CTA;
```
jnsdls, line 49: please no default exports unless it's required

components/faucet/CTA.tsx
```diff
@@ -0,0 +1,49 @@
+import { Flex } from "@chakra-ui/react";
+import Link from "next/link";
+import type { FC } from "react";
+import { Button, Text } from "tw-components";
+
+const CTA: FC = () => {
```
jnsdls, line 6:
```suggestion
export const CTA: React.FC = () => {
```

prefer named exports & `React.FC` (no need to separately import `FC` that way)

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+        description:
+          (err as any).response.data.message || "Something went wrong",
+        status: "error",
+        duration: 9000,
+        isClosable: true,
+      });
+    } finally {
+      setLoading(false);
+    }
+  };
```
jnsdls, line 38: this should be a `useMutation` (gives u the loading state etc for free)

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
+import { Input, InputGroup, Spinner, chakra, useToast } from "@chakra-ui/react";
+import axios from "axios";
+import { FC, useState } from "react";
+import { Button } from "tw-components";
+
+const FormComponent: FC = () => {
+  const toast = useToast();
+  const [loading, setLoading] = useState(false);
```
jnsdls, line 8: not needed with mutation

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
+import { Input, InputGroup, Spinner, chakra, useToast } from "@chakra-ui/react";
+import axios from "axios";
+import { FC, useState } from "react";
+import { Button } from "tw-components";
+
+const FormComponent: FC = () => {
```
jnsdls, line 6:
```suggestion
export const FormComponent: React.FC = () => {
```
same as the other component

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+        duration: 9000,
+        isClosable: true,
+      });
+    } finally {
+      setLoading(false);
+    }
+  };
+
+  return (
+    <chakra.form
```
jnsdls, line 41: wy not `<Form>` which chakra has?

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+          value={address}
+        />
+      </InputGroup>
+
+      <Button
+        bg="#0098EE"
+        borderColor="#0098EE"
+        color="#F2F2F7"
+        disabled={loading}
+        onClick={sendFunds}
```
jnsdls, line 63: since we're in a form here we should use the form `onSubmit` behavior not an onclick (otherwise there's no point to wrap it in a form in the first place)

components/faucet/FormComponent.tsx
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  return (
+    <chakra.form
+      alignItems="center"
+      display="flex"
+      gap="4"
+      justifyContent="center"
+    >
+      <InputGroup>
```
jnsdls, line 47: `<InputGroup>` only makes sense if we're grouping something, either add error states or labels or something or remove the extra component

pages/api/faucet/solana.ts
```diff
@@ -0,0 +1,29 @@
+import * as web3 from "@solana/web3.js";
```
jnsdls, line 1: prefer only importing the things u actually need

pages/api/faucet/solana.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
+  if (req.method !== "POST") {
+    return res.status(400).json({ message: "Please use POST method" });
+  }
+
+  const { address } = req.body;
+
+  try {
+    const connection = new web3.Connection("https://api.devnet.solana.com");
```
jnsdls, line 12: branch should be based off of the solana branch at which point u can read the RPC url from the config file for it instead of hard-coding it

pages/api/faucet/solana.ts
```diff
@@ -0,0 +1,29 @@
...
+    const wallet = new web3.PublicKey(address);
+
+    const txHash = await connection.requestAirdrop(
+      wallet,
+      web3.LAMPORTS_PER_SOL * Number("1"),
```
jnsdls, line 17: `Number("1")` is the same as `1`, also multiplying by `1` does nothing?

pages/faucet/solana.tsx
```diff
@@ -0,0 +1,38 @@
...
+const SolanaFaucet: NextPage = () => {
+  return (
+    <AppLayout>
+      <NextSeo
+        title="Solana faucet"
+        description="Get Solana devnet tokens for free"
+        openGraph={{
+          title: "Solana faucet | thirdweb",
+          url: `https://thirdweb.com/faucet/solana`,
+        }}
```
jnsdls, line 18: these need work, rahul should be able to help with the copy

pages/faucet/solana.tsx
```diff
@@ -0,0 +1,38 @@
...
+      />
+      <Flex flexDir="column" maxW="900px" w="full">
+        <Heading color="#F2F2F7">Solana faucet</Heading>
+        <Text
+          as="h2"
+          color="whiteAlpha.800"
+          fontSize="2xl"
+          fontWeight="bold"
+          mb="4"
+        >
```
jnsdls, line 28: based on how ur styling this this probably should be a `<Heading>` component which has most of these behaviors of styling by default

jnsdls (changes requested):
thx for this @avneesh0612! left a bunch of comments for things to clean up, the main thing is that you probably will want to base this PR on top of https://github.com/thirdweb-dev/dashboard/pull/486, which has the solana work we're doing and will make things like getting the correct RPC url easier.

pages/api/faucet/solana.ts
avneesh0612: oops yea, I was doing dynamic amount before so forgot to update

PR #520, opened 2022-09-29
Remove loading states in tabs

nachoiacovino:


jnsdls:
so this would mean the tabs would just "pop in" now?

nachoiacovino:
> so this would mean the tabs would just "pop in" now?

yeah, what do you suggest? maybe an animation?

jnsdls:
> > so this would mean the tabs would just "pop in" now?
> 
> yeah, what do you suggest? maybe an animation?

I think we have the loading states maybe just not show the text but still be there or something. something just popping in is always bad.

nachoiacovino:
Superseded by #532

PR #527, opened 2022-09-30
Ignore some more non-actionable errors

nachoiacovino:


sentry.config.js
```diff
@@ -25,6 +25,24 @@ export const sentryOptions = {
...
+    "Cannot read properties of undefined (reading 'loaded')",
+    "Cannot read properties of undefined (reading 'outputCurrentConfiguration')",
+    "apstagLOADED is not defined",
+    "moat_px is not defined",
+    "ResizeObserver loop limit exceeded",
+    "null is not an object (evaluating 'e.uid')",
+    "window.ReactNativeWebView.postMessage is not a function",
+    "_reportEvent is not defined",
+    "requestAnimationFrame is not defined",
+    "window.requestAnimationFrame is not a function",
```
jnsdls, line 43: technically these mean the user is on an outdated browser, not *not* actionable but I agree they can be ignored for now since we are not gonna go support random year old browsers atm

sentry.config.js
```diff
@@ -25,6 +25,24 @@ export const sentryOptions = {
     // Common non-actionable errors
     "rejected transaction",
     "User closed modal",
+    "Loading chunk",
+    "Failed to execute '",
+    "NetworkError when attempting to fetch resource.",
+    "googlefc is not defined",
+    "__cmp is not defined",
+    "Cannot read properties of undefined (reading 'cmp')",
+    "Cannot read properties of undefined (reading 'loaded')",
```
jnsdls, line 34: this means we would even ignore errors that happen if *we* try to access `.loaded` on something undefined so this is a bit too scary to exclude imo

sentry.config.js
```diff
@@ -25,6 +25,24 @@ export const sentryOptions = {
...
+    "apstagLOADED is not defined",
+    "moat_px is not defined",
+    "ResizeObserver loop limit exceeded",
+    "null is not an object (evaluating 'e.uid')",
+    "window.ReactNativeWebView.postMessage is not a function",
+    "_reportEvent is not defined",
+    "requestAnimationFrame is not defined",
+    "window.requestAnimationFrame is not a function",
+    "tronLink.setAddress is not a function",
+    "null is not an object (evaluating 't.uid')",
```
jnsdls, line 45: can't ignore this

jnsdls (changes requested):
mostly good, couple of things to consider that we should not filter for now I think (don't want to over-filter either

jnsdls (approved):
:shipit:

PR #543, opened 2022-10-07
fix(faucet): minor issues

avneesh0612:


components/faucet/FormComponent.tsx
```diff
@@ -1,28 +1,63 @@
...
+            label: "success",
+          });
+        }
       },
       onError: () => {
-        setError("Please try again in sometime");
+        trackEvent({
+          category: "solana-faucet",
+          action: "request-funds",
+          label: "error",
```
nachoiacovino, line 59: send the error here

onError: (error)

then pass error in the trackEvent

nachoiacovino:
Let's use the banner as the success and add the link to Solana scan in the banner itself as text (kinda tertiary button but should be a text)

PR #550, opened 2022-10-10
Fixed Broken ContractKit Links

VampishWolf:
Pushing a fix for the links, now users would not be annoyed to see a page not found error. 🌈


nachoiacovino:
Super helpful, thanks!

PR #563, opened 2022-10-12
added more events to solanathon

saminacodes:


saminacodes:
i think the text needs centering and the mobile text can probably be smaller to accommodate the longer titles if needed

PR #571, opened 2022-10-13
Add Mint Additional Supply to Solana

nachoiacovino:
Supply/Owner doesn't get updated, we need supply to get updated so we can not show owner if supply > 1


core-ui/nft-drawer/useNftDrawerTabs.tsx
```diff
@@ -58,6 +58,30 @@ export function useNFTDrawerTabs(
           ),
         },
       ];
+
+      if (contractOrProgram.accountType === "nft-collection") {
+        tabs = tabs.concat([
+          {
+            title: "Mint",
+            // TODO: Disable if the user is not the authority
+            isDisabled: false,
```
jnsdls, line 67: are we good to go live without this check?

jnsdls (approved):
LGTM, can clean it up more later

PR #585, opened 2022-10-18
Show switch between all/active listings

nachoiacovino:
Removed the isPreviousData because it will bug out if activeListings === allListings, and it's only relevant when listings > 50, but tell me if I missed something and there's an obvious way of bringing it back @jnsdls


contract-ui/tabs/listings/components/table.tsx
```diff
@@ -193,28 +221,6 @@ export const ListingsTable: React.FC<ListingsTableProps> = ({ contract }) => {
                 </Tr>
               );
             })}
-            {getAllQueryResult.isPreviousData && (
```
jnsdls, line 196: can we instead keep this but also check if the filter didn't change? to do that use a `usePrevious()` (we have it in other places) and check it here, too before showing the overlay

jnsdls (changes requested):
we can do this better

jnsdls (approved):
LGTM

PR #586, opened 2022-10-18
[WIP] Add file and JSON upload for URI field in contract explorer

adam-maj:


components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [data, setData] = useState<any>("");
+
+  const handleUpload = async () => {
+    const [uri] = await upload(data);
+    setValue(uri);
+    onClose();
+  };
+
+  return (
+    <Drawer
```
jnsdls, line 73: why does this need a full drawer?

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,108 @@
...
+  setValue: (value: string) => void;
+}
+
+const UploadDrawer: React.FC<UploadDrawerProps> = ({
+  isOpen,
+  onClose,
+  setValue,
+}) => {
+  const { mutateAsync: upload } = useStorageUpload();
+  const [data, setData] = useState<any>("");
```
jnsdls, line 64: pls no any

jnsdls (changes requested):
I think we can improve this quite a bit

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,108 @@
...
+  onClose: () => void;
+  setValue: (value: string) => void;
+}
+
+const UploadDrawer: React.FC<UploadDrawerProps> = ({
+  isOpen,
+  onClose,
+  setValue,
+}) => {
+  const { mutateAsync: upload } = useStorageUpload();
```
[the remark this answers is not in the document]
adam-maj: The PR is WIP rn ser

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,108 @@
...
+          borderRadius="md"
+          width="200px"
+          height="200px"
+          border="1px solid"
+          borderColor="gray.500"
+          bg="gray.700"
+        >
+          <Icon as={FiUpload} color="gray.600" _hover={{ color: "gray.500" }} />
+        </Center>
+      </FileInput>
```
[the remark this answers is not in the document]
adam-maj: The base input is a string input for URI -> then it has upload button on the right, this can upload a drawer (but I'm thinking accordion is actually better for this and will replace) -> this has option to upload JSON metadata, or file.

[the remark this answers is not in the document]
jnsdls: I'm sold on the idea of "if it includes `uri` it should allow a file upload"

I am much less sold on the idea of "if it includes `uri` it should open a whole separate ui that specifically allows u to input json / upload a file"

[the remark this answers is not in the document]
adam-maj: That was the original idea - but since you can upload `json` as a file, maybe its best to just support the simple case of file upload for now.

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+    setValue(uri);
+  };
+
+  return (
+    <>
+      <InputGroup>
+        <Input
+          defaultValue={value}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
```
jnsdls, line 29: while something is uploading I should not be allowed to enter anything into this input

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+}) => {
+  const { mutateAsync: upload, isLoading } = useStorageUpload();
+
+  const handleUpload = async (file: any) => {
+    const [uri] = await upload({ data: [file] });
+    setValue(uri);
+  };
+
+  return (
+    <>
```
jnsdls, line 24: fragment is no longer used

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+  setValue: (value: string) => void;
+}
+
+export const ContractUriUpload: React.FC<ContractUriUploadProps> = ({
+  value,
+  setValue,
+}) => {
+  const { mutateAsync: upload, isLoading } = useStorageUpload();
+
+  const handleUpload = async (file: any) => {
```
jnsdls, line 18: feel free to type this as `File`

jnsdls: actually do we even need this? why can't it just be the fn on file input directly?

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+interface ContractUriUploadProps {
+  value: string;
+  setValue: (value: string) => void;
+}
+
+export const ContractUriUpload: React.FC<ContractUriUploadProps> = ({
+  value,
+  setValue,
+}) => {
+  const { mutateAsync: upload, isLoading } = useStorageUpload();
```
jnsdls, line 16: if u use mutateAsync that means you *have* to `try{}catch{}`it (it will throw if it fails)

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+          defaultValue={value}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+        />
+        <InputRightElement width="96px">
+          <FileInput setValue={(file) => handleUpload(file)}>
+            <Button
+              size="xs"
+              padding="10px"
+              paddingY="14px"
```
jnsdls, line 36: can we use chakra prebuilt values for this instead of explicit px?

components/contract-functions/contract-uri-upload.tsx
```diff
@@ -0,0 +1,48 @@
...
+    <>
+      <InputGroup>
+        <Input
+          defaultValue={value}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+        />
+        <InputRightElement width="96px">
+          <FileInput setValue={(file) => handleUpload(file)}>
+            <Button
```
jnsdls, line 33: wait in general can this not just be an icon button?

jnsdls (changes requested):
some more things to look at / clean up

components/contract-functions/contract-uri-upload.tsx
adam-maj: I prefer having the explicit upload text + styling gets messed up with icon button

jnsdls (approved):
ship it

PR #597, opened 2022-10-20
Add Web3Button option to code tab

nachoiacovino:
This feels good enough for now, we _could_ have the Web3Button appear below the code, but will need passing more props to the CodeSegment component which I don't think it's worth.

I think this helps with discovery and the link to doc completes the flow.


jnsdls (approved):
LGTM - can be improved further but it's a start

PR #604, opened 2022-10-24
useEffect for setValue

nachoiacovino:
I hate this but Chakra changed how useClipboard works in this PR https://github.com/chakra-ui/chakra-ui/pull/6763/files, basically making it so you now have setValue, for some reason it doesn't take refresh on rerenders now.

I also added a basic isValidURL function to avoid SDK errors when writing the openzeppelin relayer and a useMemo to the embedCode gen


jnsdls (approved):
change makes sense, ship it

PR #614, opened 2022-10-26
Add linting rule for no console.logs

nachoiacovino:
thoughts @jnsdls ?


jnsdls:
if we don't want console logs we can also just remove them during compilation...

jnsdls (approved):
LGTM

PR #626, opened 2022-11-04
Updated metadescription

juandolealt:
Google didn't take our metadesc because I wrote "build, launch, and manage web3 apps..." - needs to include copy on the homepage so changed this to "build web3 apps easily" & will be more likely to use it.

"Build, launch, manage" is better messaging & we can switch back to it once we update homepage copy (there have been branding discussions here)


jnsdls (approved):
LGTM

nachoiacovino (approved):
makes sense!

PR #638, opened 2022-11-09
Add OSRoyaltyToPrebuilt mapping

nachoiacovino:
thoughts on implementation @jnsdls ?

Will also like to testjam it


nachoiacovino:
We're not disabling chains correctly and only allowing to deploy from Ethereum/Goerli as we should (it's checking the allowed chains for the prebuilts), will look into best way of fixing

nachoiacovino:
Ended up hardcoding it rather than moving all the logic from custom contract form to prebuilt contract form, lmk your thoughs

jnsdls (approved):
LGTM for the short term

PR #646, opened 2022-11-15
Add displayName/description for contractParams

nachoiacovino:


jnsdls (approved):
LGTM - not worried about this breaking anything and we can always ship improvements

PR #652, opened 2022-11-17
feat: init faucet

avneesh0612:


.env.example
```diff
@@ -12,4 +12,5 @@ NEXT_PUBLIC_RPC_AVALANCHE=
 NEXT_PUBLIC_RPC_POLYGON=
 NEXT_PUBLIC_RPC_MUMBAI=
 FEEDBACK_WEBHOOK_URL=
-SSR_ALCHEMY_KEY=
\ No newline at end of file
+SSR_ALCHEMY_KEY=
+PRIVATE_KEY=
```
jnsdls, line 16: definitely not putting private key in env vars

components/app-layouts/providers.tsx
```diff
@@ -65,6 +65,11 @@ export const DashboardThirdwebProvider: ComponentWithChildren = ({
       }}
       storageInterface={StorageSingleton}
       walletConnectors={walletConnectors}
+      authConfig={{
+        authUrl: "/api/auth",
+        domain: "thirdweb.com",
+        loginRedirect: "/faucet/mumbai",
+      }}
```
jnsdls, line 72: auth is a whole different discussion I don't think we're ready to implement auth on the dashboard yet

pages/api/faucet/mumbai.ts
```diff
@@ -0,0 +1,81 @@
...
+    method: "alchemy_getAssetTransfers",
+    params: [
+      {
+        fromBlock: "0x0",
+        fromAddress: "0xdf62A3fb98B160D98b406F055EdE42FB26C7327B",
+        toAddress: address,
+        category: ["external"],
+      },
+    ],
+  });
```
jnsdls, line 30: this will not work once we switch to our own public RPC (soon) because it's an alchemy specific call (alchemy_ prefix) and it is not guaranteed to support that

pages/api/faucet/mumbai.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (req.method !== "POST") {
+    return res.status(400).json({ message: "Please use POST request" });
+  }
+
+  const { address } = JSON.parse(req.body);
+
+  const wallet = new ethers.Wallet(
+    process.env.PRIVATE_KEY as string,
+    new ethers.providers.JsonRpcProvider(EVM_RPC_URL_MAP[ChainId.Mumbai]),
+  );
```
jnsdls, line 16: it is much more likely that we would set up a faucet as a smart contract than as a wallet

jnsdls (changes requested):
this is nice in theory but I don't think the approach fits the way we would implement this type of faucet

jnsdls:
we'll attack this differently

PR #662, opened 2022-11-22
add support for default values for releases

jnsdls:


jnsdls:
actually this isn't blocked...

PR #678, opened 2022-11-28
Show cc for sig drop with only specific wallets

nachoiacovino:


nachoiacovino:
my bad this is not necessary but something weird going on, not this :D

PR #680, opened 2022-11-29
Improve marketplace drawer

nachoiacovino:
We don't currently return token standard for listing asset, what we call "type" is listing type (direct or auction), fixed it to show that and also fixed not showing owner for erc1155 on general NFT drawer which was pending


jnsdls (approved):
LGTM

PR #682, opened 2022-11-29
GamingKit landing page

nachoiacovino:

PR #693, opened 2022-12-02
Refactor pages/index.tsx

MananTank:
## What?

* Break down `pages/index.tsx` into subcomponents by creating a separate component for each section
* No Additional code change is done 

## Why?

* `pages/index.tsx` is a huge file (850+ LOC) and is very hard to maintain
* Breaking it down to subcomponents makes it easy to maintain 
* it will also allow us to load some of the sections lazily using `dynamic()` to reduce initial JS bundle size and improve main thread blocking during page load


pages/index.tsx
```diff
@@ -75,679 +21,7 @@ const HomePage: ThirdwebNextPage = () => {
...
-              <GeneralCta
-                title="Start building for free"
-                size="lg"
-                w={{ base: "full", md: "inherit" }}
-              />
-            </LightMode>
-          </Flex>
-        </HomepageSection>
-
+        <HomePageSections />
```
jnsdls, line 24: I think I would still like the sections to be here in-line, we can do the dynamic import in this file, too.

Reason being as we add and remove sections / move them around it's not an extra step to first go to the sections file (that doesn't really do anything but to render them anyways)

jnsdls (commented):
thanks for this, been meaning to get to this for a long time

jnsdls:
anything below the newsletter signup section can probably be wrapped in suspense / lazy loaded (should be enough offset to be OK)

jnsdls:
@MananTank if u want to try to apply this method within the sections that can be lazy loaded: https://vercel.com/blog/improving-interaction-to-next-paint-with-react-18-and-suspense#case-study-next.js-site

pages/index.tsx
MananTank: @jnsdls Ok, Let me Delete `<HomePageSections/>` and move all sections to `pages/index.tsx`

MananTank:
@jnsdls I've Deleted `<HomePageSections/>` and moved all sections to `pages/index.tsx` according to the suggestion - Please verify

We can create a new issue & branch for improving the Hydration perf.

nachoiacovino:
one thing I see is that ExamplesSection is not really reusable right now and we are using that in /auth, can you make it so either we reuse that or we make it easy to reuse it if we want to refactor later? (basically one uses HomepageSection and the other uses ProductSection), also if you check my branch #682, I already made the coding section reusable so this might clash

MananTank:
@nachoiacovino Can you clarify which sections are you referring to as ExamplesSection and Coding Section?

nachoiacovino:
so examples section is the one with 8 examples with link to github, we have a very similar one in auth, and the code section is the one with code examples and link to replit that we're gonna be reusing in in a new landing page, you can see in #682

jnsdls:
@nachoiacovino I think we can do the refactor for more re-usable examples and code sections separately - or do you think we need it as part of this PR?

nachoiacovino (approved):
no it's fine, we're just reusing CodeSelector which doesn't have any new required props so we'll be fine there!

I'll be merging this, if you want to take next steps, I think making ExampleSection reusable to the landing version and auth would be great, thanks for the contribution!

PR #721, opened 2022-12-10
optimize-newsletter: drop react-hook-form

MananTank:
## Why?

The newsletter form only has a single input element as shown below

image

* using `react-hook-form` for handling a single input is not necessary 
* removing `react-hook-form` from the newsletter will reduce JS bundle size by a 22kB on any page that use it (unless of course, that page has other components using `react-hook-form` )

### Before ( landing page )
image

image

### After ( landing page ) - 22kB JS reduction

image

---

## Code changes 

* `react-hook-form` replaced with idiomatic code for handling submission - Yes, Its a bit verbose but reduces a lot of code in the final bundle size
* functionality remains the same ( prevent submission on invalid email, preventDefault for submit event, clear input after submission, loading state indicator )

<br/>

### DIFF
--- a/components/homepage/sections/NewsletterSection.tsx
+++ b/components/homepage/sections/NewsletterSection.tsx
@@ -7,16 +7,35 @@ import {
   Input,
   Stack,
 } from "@chakra-ui/react";
-import { useForm } from "react-hook-form";
+import { useState } from "react";
 import { MdMarkEmailRead } from "react-icons/md";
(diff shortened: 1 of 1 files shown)


jnsdls:
I think this could even be done with an uncontrolled input via ref instead (or even via the native behavior of `<form>` without requiring any state hook at all. with that said this seems like a valid improvement, so will merge it as is once visually checked over in a minute :) thx @MananTank!

PR #733, opened 2022-12-13
cleanup .env file and add description to each env var

### LINKED ISSUE #728: [Question] How can I get the values for the env variables?
I couldn't find documentation on how to get the values for the environment variables. 

If it is documented somewhere, Can you please give me the link to it? ( It would be a good idea to include that in the README ) and If it is not documented - Can you add it to the README?

I have been using placeholder values for the environment variables but I'm having trouble with it when running the `next build` and I would like to properly set up my local environment so that I can build pages properly

jnsdls:
fixes #728


jnsdls:
#### 📝 Changed routes:




1 deleted route:
- **/api/feedback**

Commit 24c05ea4bc9c4409aab44f7f3ab8df2f734488da (https://thirdweb-in4a6z6b6.thirdweb-preview.com).

PR #744, opened 2022-12-20
Fix emotion's string not allowed as css prop error in UpdateNotice

MananTank:
PR fixes the below-shown error thrown by [emotion](https://github.com/emotion-js/emotion) in dev when clicking on the "Claim conditions" tab as shown below:

image


jnsdls (approved):
LGTM

PR #745, opened 2022-12-20
Refactor claim-conditions.tsx

MananTank:
### Why 
Cleaning up `claim-conditions.tsx` to prepare for [FID-1279](https://linear.app/thirdweb/issue/FTD-1279/claim-condition-saving-is-acting-erratic) 

### Code Change
Moved components `QuantityInputWithUnlimited`, `PriceInput` and `ClaimConditionsForm` to separate files to make the `claim-conditions.tsx` more maintainable and easier to work with. No additional changes

PR #746, opened 2022-12-21
Fix file name of QuantityInputWithUnlimited component

MananTank:
Fix the wrong file name introduced in https://github.com/thirdweb-dev/dashboard/pull/745 as pointed out by @nachoiacovino in https://github.com/thirdweb-dev/dashboard/pull/745#issuecomment-1361160477 

Changing `quality-input-with-unlimited.tsx` to `quantity-input-with-unlimited.tsx` which contains the `QuantityInputWithUnlimited` component


nachoiacovino:
hey @MananTank something I just realized. 

the component that are imported both on contracts and programs, cannot live under contract-ui, since it doesn't belong only in evm, and it belongs to both evm and solana, it needs to live under core-ui. 

Sorry I missed it before!

MananTank:
@nachoiacovino Yeah, I've seen a lot of such cases. Many reusable components are put in a hierarchy that doesn't really make sense in hindsight

We should probably create an issue and list down all such instances and refactor all at once

What do you think?

Issue #747, opened 2022-12-21
/sdk page has a Hydration Error


MananTank:
## Hydration Error

<img width="987" alt="Pasted Graphic" src="https://user-images.githubusercontent.com/22043396/208905036-ca40feaa-e5f2-4b5d-a8cd-35ea35c78cdd.png">

<img width="738" alt="Pasted Graphic 1" src="https://user-images.githubusercontent.com/22043396/208905123-4f22e924-d9cd-4f11-a8ce-2256054f16c9.png">

### Root Cause 

Icons container `<div>` is put inside a `<p>` - which is invalid HTML nesting and thus creates an issue with Hydration

<img width="500" alt="image" src="https://user-images.githubusercontent.com/22043396/208905533-bf1c5121-6709-4afb-a593-992bdd8cbfda.png">

<img width="915" alt="image" src="https://user-images.githubusercontent.com/22043396/208905823-718df310-0f32-47d3-9671-b2cbfc701f81.png">

PR #748, opened 2022-12-21
Fix Hydration Error in /sdk page by fixing ProductCard

MananTank:
Fixes https://github.com/thirdweb-dev/dashboard/issues/747 

## Code Change

* Change `<p>` to `<div>` by Using `<Box>` instead of `<Text>` as a container of `props.children` for `ProductCard` component to fix hydration error in `/sdk` page and future hydration errors as well
* Added `color` and `lineHeight` styles to maintain styles as it was earlier

### DIFF
--- a/components/product-pages/common/ProductCard.tsx
+++ b/components/product-pages/common/ProductCard.tsx
@@ -1,6 +1,6 @@
-import { Flex } from "@chakra-ui/react";
+import { Box, Flex } from "@chakra-ui/react";
 import NextImage, { StaticImageData } from "next/image";
-import { Heading, Text } from "tw-components";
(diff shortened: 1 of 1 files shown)


nachoiacovino:
Seems like it was broken from before but just checked on mobile and this is how it looks image

Probably would be good to fix it on this same PR.

MananTank:
@nachoiacovino Yeah, It was broken already. I just fixed the Hydration error - and kept the UI as is. 

Since this branch is already merged, Let me fix that in a new branch

MananTank:
The UI issue fixed in https://github.com/thirdweb-dev/dashboard/pull/749

PR #749, opened 2022-12-21
SDK page: Fix icon positions in ProductCard

MananTank:
## The Problem

* icons in ProductCard the `/sdk` page are not properly positioned in Mobile View
image

* Their position on Desktop also gets distorted when content is increased 

image

<br />

## The Solution

PR Fixes above mentioned two issues as shown below:

image

image


nachoiacovino (approved):
awesome! lgtm

PR #754, opened 2022-12-23
feat: make the events page

avneesh0612:


components/events/Event.tsx
```diff
@@ -0,0 +1,134 @@
...
+          }}
+        >
+          {type}
+        </Badge>
+        <Flex flexDir="column" w="full">
+          <Heading textAlign="left" size="title.md">
+            {title}
+          </Heading>
+
+          <Flex justify="space-between" mt={2}>
```
shift4id, line 67: This spacing looks a bit weird, try it with a gap between the items and justify at the start.

components/events/Event.tsx
```diff
@@ -0,0 +1,134 @@
...
+            <Flex align="center" gap={1}>
```
shift4id, line 68: text breaks when the screen is small, so add flex wrap to ensure each individual text stays on one line

components/events/FeaturedCard.tsx
```diff
@@ -0,0 +1,75 @@
+import { Flex, Icon, Image } from "@chakra-ui/react";
+import { useTrack } from "hooks/analytics/useTrack";
+import type { FC } from "react";
```
shift4id, line 3: Don't need to import FC, can just use React.FC<>

components/events/Event.tsx
```diff
@@ -0,0 +1,134 @@
...
+              category: `events-${title}`,
+              action: "click",
+              label: "register",
+            })
+          }
+          w={60}
+          fontSize="20px"
+          color="black"
+          flexShrink={0}
+          background="rgba(255,255,255,1)"
```
shift4id, line 123: should be able to use white for background and whiteAlpha.900 for hover background. tested and works on local.

components/events/Event.tsx
```diff
@@ -0,0 +1,134 @@
...
+            </Flex>
+
+            <Flex align="center" gap={1}>
+              <Icon as={FaMapMarkerAlt} />
+              <Text>{location}</Text>
+            </Flex>
+          </Flex>
+        </Flex>
+      </AccordionButton>
+      <AccordionPanel
```
shift4id, line 97: The panel content is misaligned with the button

shift4id (changes requested):

avneesh0612:
Just re-requested request and it removed nacho from reviewer 🤦 . My bad.

nachoiacovino (changes requested):
oh I had a review done like a week ago and I forgot to submit it 💀

nachoiacovino:
more spacing everywhere here 
image

nachoiacovino:
this text gets cut out 
image

nachoiacovino:
image

this is all over the place with alignments, it doesn't look clean, also it should have some sort of hover interaction

nachoiacovino:
this also needs work on mobile 
image

avneesh0612:
implemented stuff from both reviews. cc: @nachoiacovino @shift4id 🫡

components/devRelEvents/FeaturedCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <Flex flexDir="column" p={4} align="center">
+        <Heading>{title}</Heading>
+        <Flex align="center" my={2}>
+          <Icon as={FiCalendar} color="gray.300" />
+          <Text ml={2} color="gray.300">
+            {date}
+          </Text>
+        </Flex>
+
+        <LinkButton
```
shift4id, line 52: Give this button a little more margin with the bottom of the card, it looks a little dense

pages/events.tsx
```diff
@@ -0,0 +1,231 @@
...
+                  date={date}
+                  banner={banner}
+                  link={link}
+                />
+              ))}
+            </Grid>
+          </HomepageSection>
+
+          <HomepageSection mt={20}>
+            <Select
```
shift4id, line 180: Try moving it to the right and reducing the bottom margin because it's equidistant from both the featured cards and events list. If moving it to the right doesn't look good then keep it in place.

shift4id (changes requested):

pages/events.tsx
```diff
@@ -0,0 +1,231 @@
...
+const EventsPage: ThirdwebNextPage = () => {
+  const [sortBy, setSortBy] = useState<"date" | "location" | "type" | "title">(
+    "date",
+  );
+  const [sortedEvents, setSortedEvents] = useState(allEvents);
+
+  const handleSort = (value: "date" | "location" | "type" | "title") => {
+    setSortBy(value);
+    switch (value) {
+      case "date":
```
jnsdls, line 106: date should probably be the default filter

filtering by title makes little sense to me

jnsdls: instead of having filters we can always sort by title and have a "search" box in addition that filters by location, type and title based on string matching?

components/devRelEvents/DevRelEvent.tsx
```diff
@@ -0,0 +1,149 @@
...
+  title,
+  timestamp,
+  location,
+  description,
+  link,
+}) => {
+  const trackEvent = useTrack();
+
+  return (
+    <AccordionItem
```
jnsdls, line 33: it's not clear that this is clickable, if anything let's make the entire card clickable (accordionbutton can have the padding instead of the item and then everything will be clickable)

components/devRelEvents/DevRelEvent.tsx
```diff
@@ -0,0 +1,149 @@
...
+        gap={4}
+        ml={{
+          base: 0,
+          md: "8.5rem",
+        }}
+        justifyContent="left"
+      >
+        <Text textAlign="left">{description}</Text>
+        <LinkButton
+          href={link}
```
jnsdls, line 127: this needs to open in a new tab if it links out to a non-thirdweb.com site please

jnsdls (changes requested):
couple of changes that I found

pages/events.tsx
avneesh0612: so it would search for everything? 🤔

avneesh0612: Maybe we could do like a search using the title and add sort using location, type, etc.?

jnsdls: what's the sorting parameter for "location"? how close the event is to the user? alphabetical makes no sense, that's not sort but grouping then.
same question for type

jnsdls: "search" everything means the user can type "san francisco" and they see everything in san francisco, which makes sense, or they can type "hackathon" and they'll see all hackathons, or they can search a specific title and they'll find that, aka it gets users to success faster

shift4id (approved):

pages/events.tsx
avneesh0612: hmm yea that makes sense

avneesh0612: so search just using the title?

pages/events.tsx
```diff
@@ -0,0 +1,205 @@
...
+  {
+    type: "Hackathon",
+    title: "Ready Player 3 Hackathon",
+    date: "Jan 16th - 31st",
+    banner: "/assets/og-image/readyplayer3.png",
+    link: "https://thirdweb.com/hackathon/readyplayer3",
+  },
+];
+
+const allEvents = [
```
jnsdls, line 23: some of these are now already in the past => we'll need a toggle between "upcoming" and "past"

jnsdls (changes requested):

pages/events.tsx
avneesh0612: hmm a dropdown?

jnsdls: a select should do it, default value should be "upcoming"

jnsdls (approved):
good to ship

PR #763, opened 2022-12-29
Support network chain-id path in Dashboard

MananTank:
Currently dashboard uses the `/<network-name>/<contract_address>` path scheme

Add support for `/<chain_id>/<contract_address>` as well 

For example: `/80001/<contract-address>` should open the same page as `/mumbai/<contract-address>` ( mumbai's chain id is 80001 )


jnsdls (approved):
LGTM thx!

Issue #758, opened 2022-12-27
Random behavior with text rendering


MananTank:
Usually, on the first load ( and sometimes even after that ) - the website is not using the "Inter" font-family and uses the fallback "sans-serif" to render the text. But when we reload the page - the issue gets resolved.

To replicate this issue, Load the website in incognito, and check the font-family:

<img width="1275" alt="image" src="https://user-images.githubusercontent.com/22043396/209622533-17ef9500-cf98-4a67-a92c-104573646e2a.png">

Then reload the page, and check again:

<img width="1229" alt="image" src="https://user-images.githubusercontent.com/22043396/209622860-79824b8f-d603-4b8a-a0df-94da3baae5ec.png">

jnsdls:
I believe this is a @next/font issue, if that's the case others will be running into it, too

PR #765, opened 2023-01-02
Fix font loading issue

MananTank:
Fixes https://github.com/thirdweb-dev/dashboard/issues/758


MananTank:
Root cause of issue seems to be `font-display: optional;` used on the @ font-face by default by Next.js. 

Setting it "optional" means that "Gives the font face an extremely small block period and no swap period." - Changing it to "swap" seems to fix the issue

More info:
https://nextjs.org/docs/api-reference/next/font#display
https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display#values

PR #767, opened 2023-01-03
Possible solution to the large bundle size caused by AddressCopyButton

JustinTime42:
I noticed this comment in AddressCopyButton: 

```
/**
 * 📝 @TODO
 * Write a custom alternative to `shortenIfAddress` function
 * Why? - it imports a lot of other stuff which can not be tree-shaken
 */
import { shortenIfAddress } from "utils/usedapp-external";

/**
 * ❗️❗️❗️ @DANGER
 * Importing this component will result in a HUGE bundle size because of `shortenIfAddress`
 */
```

The shortenIfAddress function imports Ethereum and Solana libraries to see if the address passed to the button is an EVM or Solana address. If it is either one, it shortens the address to the first 6, and last 4 characters. Otherwise it leaves the address untouched. I wonder, for this button component, if maybe we don't care what sort of address it is, and if we might benefit from being able to support shortening for other kinds of obnoxiously long strings such as long token IDs or IPFS addresses. 

This pull request replaces all of those imports and logic with a simple function to just shorten whatever string is passed to the first 6, and last 4 characters unless the shortenAddress prop is passed as false (or unless the string is already short). A glance through the various places where the button is currently implemented appears to work fine with what I've changed. However, it's entirely possible that this extremely simplified approach misses edge cases or future features that I'm not aware of. 

Still, I thought it'd be a good place to start and hopefully wind up with something that allows us to get rid of those massive imports.


jnsdls (commented):
In its current state this PR means the address copy button no longer checks if it is being passed an address and will always shorten anything passed.

I am not sure *yet* whether that is an acceptable trade-off, maybe there is a way to keep the behavior the same while also no longer relying on the `ethers` and `metaplex/js` packages...

alternatively it would probably be acceptable to make the button in itself more generic as to accept any string (potentially even as `children`?) and provide a shortened length (optionally) to shorten the displayed contents.

@MananTank I would love t hear your ideas on this, too

jnsdls:
oh and also thanks for the PR @JustinTime42! 👍

jnsdls:
@MananTank I'm happy with that reasoning and decisions and will leave it to you and @JustinTime42 to get this over the finish line :)

MananTank:
@jnsdls @JustinTime42 

With this code change - I'm not seeing JS bundle size improvement on any page 😭

It seems that pages using the AddressCopyButton are also importing the web3 libraries that we were trying to shake off 😅 

I think this code change should still be merged so that If in the future we build a page where we need `<AddressCopyButton/>` but don't import web3 libraries - it doesn't have excess bundle size

image

PR #770, opened 2023-01-04
Homepage UI overhaul - Dark Variant

MananTank:


components/homepage/sections/KeyFeatures.tsx
```diff
@@ -0,0 +1,154 @@
...
+  img: string;
+}
+
+const features: FeatureCardProps[] = [
+  {
+    title: "Build.",
+    description: (
+      <>
+        Build your own contracts with{" "}
+        <Link
```
nachoiacovino, line 19: would be great if these would be `<TrackedLink>` instead of `<Link>`

MananTank: Fixed in https://github.com/thirdweb-dev/dashboard/pull/770/commits/9716a32ccb6489b24fc6f8462beaee083d09f887

PR #778, opened 2023-01-09
disable refetching on window focus

MananTank:
Fixes https://linear.app/thirdweb/issue/FTD-1279/claim-condition-saving-is-acting-erratic

### DIFF
--- a/components/app-layouts/app.tsx
+++ b/components/app-layouts/app.tsx
@@ -88,6 +88,7 @@ export const AppLayout: ComponentWithChildren<AppLayoutProps> = (props) => {
       new QueryClient({
         defaultOptions: {
           queries: {
+            refetchOnWindowFocus: false,
             // 24 hours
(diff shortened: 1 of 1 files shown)


components/app-layouts/app.tsx
```diff
@@ -88,6 +88,7 @@ export const AppLayout: ComponentWithChildren<AppLayoutProps> = (props) => {
       new QueryClient({
         defaultOptions: {
           queries: {
+            refetchOnWindowFocus: false,
```
jnsdls, line 91: this means *none* of our queries will ever re-fetch when window re-focusses. that might be OK but affects more than just the one spot so let's not take this decision without considering it first (not sure if you have already)

MananTank: No longer need this - found the proper way to fix the issue. Closing this

PR #779, opened 2023-01-09
feat: make some updates to the rp3 page

avneesh0612:


components/hackathon/common/AvatarShowcase.tsx
```diff
@@ -20,13 +20,12 @@ export const AvatarShowcase: React.FC<AvatarShowcaseProps> = ({
   return (
     <VStack mb={20} spacing={12}>
       <Heading size="title.2xl">{title}</Heading>
-      <SimpleGrid
```
nachoiacovino, line 23: why? these makes the columns not the same size
image

nachoiacovino (changes requested):

components/hackathon/common/AvatarShowcase.tsx
avneesh0612: we won't be able to center it otherwise, right?

nachoiacovino: is that super relevant? I very much prefer it to be aligned than centered

nachoiacovino (approved):
lgtm!

PR #780, opened 2023-01-09
feat: improve the winners section

avneesh0612:


components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+                boxShadow: "0 0 10px 0 rgba(0, 0, 0, 0.2)",
+                transform: "translateY(-5px)",
+                transition: "all 0.3s ease",
+              }}
+              zIndex={2}
+              key={name}
+              direction="column"
+              pos="relative"
+              bg="whiteAlpha.100"
+              w="full"
```
shift4id, line 56: w="full" is not valid, it should be w="100%"

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+              _hover={{
+                boxShadow: "0 0 10px 0 rgba(0, 0, 0, 0.2)",
+                transform: "translateY(-5px)",
+                transition: "all 0.3s ease",
+              }}
+              zIndex={2}
+              key={name}
+              direction="column"
+              pos="relative"
+              bg="whiteAlpha.100"
```
shift4id, line 55: this can be bg="white"

shift4id (changes requested):

components/hackathon/common/HackathonWinners.tsx
nachoiacovino: both are valid

components/hackathon/common/HackathonWinners.tsx
nachoiacovino: no, whiteAlpha works better (check the preview)

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
+import { Flex, Grid, Icon, Image, Link, VStack } from "@chakra-ui/react";
+import { FiExternalLink } from "react-icons/fi";
+import { Heading, Text } from "tw-components";
+
+interface IWinner {
```
nachoiacovino, line 5:
```suggestion
interface Winner {
```

We don't use the I thing anywhere (or we try not to)

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+    <VStack
+      mb={{
+        base: 40,
+        md: 0,
+      }}
+    >
+      <Heading size="title.2xl" textAlign="center" paddingTop={10}>
+        Hackathon Winners
+      </Heading>
+      <Grid
```
nachoiacovino, line 29: better if this is a SimpleGrid, otherwise we get this different sized cards
image

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+              zIndex={2}
+              key={name}
+              direction="column"
+              pos="relative"
+              bg="whiteAlpha.100"
+              w="full"
+              pb={4}
+              rounded="lg"
+            >
+              <Image
```
nachoiacovino, line 60: Can we change to ChakraNextImage?

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+                src={image}
+                alt={name}
+                objectFit="cover"
+                h={60}
+                roundedTop="lg"
+              />
+              <Flex flexDir="column" pl={4}>
+                <Heading size="title.md" color="white" mt={5}>
```
nachoiacovino, line 68: better for the margin to be on the Flex parent component rather than the Heading

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+                <Text as="h4" fontSize="md" color="gray.200">
+                  {position}
+                </Text>
+              </Flex>
+              <Icon
+                as={FiExternalLink}
+                width={6}
+                height={6}
+                color="white"
+                pos="absolute"
```
nachoiacovino, line 80: hmm why do you need the button to be positioned absolute? much easier to do a flex with justify-content space-between and some padding

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+        placeItems="center"
+        gap={10}
+        px={{
+          base: 10,
+          md: 20,
+        }}
+        py={10}
+      >
+        {winners.map(({ name, image, link, position }) => (
+          <Link href={link} isExternal key={name}>
```
nachoiacovino, line 44: I wouldn't put a Flex and all that thing inside a Link component, you should use a LinkOverlay

components/hackathon/common/HackathonWinners.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+interface HackathonWinnersProps {
+  winners: IWinner[];
+}
+
+export const HackathonWinners: React.FC<HackathonWinnersProps> = ({
+  winners,
+}) => {
+  return (
+    <VStack
```
nachoiacovino, line 20: Can we use a <Flex flexDir="column" /> with some gap instead of a VStack?

nachoiacovino (changes requested):

components/hackathon/common/HackathonWinners.tsx
avneesh0612: my bad, old habit 🥲

components/hackathon/common/HackathonWinners.tsx
avneesh0612: a?

nachoiacovino: lmao, Flex

avneesh0612: ah gotcha

jnsdls:
problem with some viewport sizes
image

section gradient is overlapping the footer (it should not)
image

avneesh0612:
what's the problem with the viewport sizes?

jnsdls:
> what's the problem with the viewport sizes?

2 on top and 1 below does not make sense to me for top 3, we should either show them next to each other or fully stacked

avneesh0612:
oh I see okay

avneesh0612:
fixed that

nachoiacovino (approved):
lgtm

PR #784, opened 2023-01-10
remove deprecated og pages

jnsdls:
this finally removes all "og-image.thirdweb.com" references so we can turn off the service - these have already not been in active use anymore for a while so it should be safe


jnsdls:
#### 📝 Changed routes:




2 deleted routes:
- **/_og/profile**
- **/_og/release**

Commit f98909ce71143b901450926f5a2878255ae10418 (https://thirdweb-8vmfrc3su.thirdweb-preview.com).

PR #789, opened 2023-01-12
Refactor claim conditions form + Fix weird behavior

MananTank:
## Issues Fixed

Linear - https://linear.app/thirdweb/issue/FTD-1279/claim-condition-saving-is-acting-erratic

* Snapshot Upload panel gets closed when uploading the snapshot (CSV/JSON file)
* Form loses all the user input when a snapshot is uploaded
* Form is not disabled when a transaction is going
* Form shows incorrect info (old data instead of user input) after submitting the form and when the transaction goes through - then it shows the correct data
* Form is not disabled when the data for it is being fetched and shows the stale data
* No spinning indicator when the data for the form is being fetched
 

## Refactor Details

* break down huge form file into components
* Each component represents a single-form field
* to avoid passing too many props - a context is created and consumed in the form fields
* repetitive form control and form group structures are extracted to single components - `<CustomFormControl/>` and `<CustomFormGroup/>` - which allows changing the UI by updating a single component instead of having to do that for all form fields separately

## Root cause of weird behavior in Claim Conditions Form
`<ClaimConditions />` ( and all other tabs ) were being unmounted when the form was either submitted or the focus moved away from the window and came back because of query re-fetching in the parent component - This resets its state and creates all sorts of UI issues

This was happening because of how it (and all other tabs) were created from `useNFTDrawerTabs()` hook and rendered in `<NFTDrawer />`

With [this](https://github.com/thirdweb-dev/dashboard/pull/789/files#diff-085e2765263ad758a1dae048567415b192dc4e4bc14239c7fa69240804375326) code change the issue gets resolved

Not only this fixes the issue in `<ClaimConditions />` but all other tabs 🎉


MananTank:
### Root cause of weird behavior in Claims Condition Form

`<ClaimConditions />` ( and all other tabs ) was being unmounted when the form was either submitted or the focus moves away from the window and comes back because of query re-fetching  - This resets its state of it and creates all sorts of UI issues

This was happening because of how it (and all other tabs) were created from `useNFTDrawerTabs()` hook and rendered in `<NFTDrawer />`

With [this](https://github.com/thirdweb-dev/dashboard/pull/789/files#diff-085e2765263ad758a1dae048567415b192dc4e4bc14239c7fa69240804375326) code change the issue gets resolved

Not only this fixes the issue in `<ClaimConditions />` but all other tabs 🎉

jnsdls (approved):
LGTM

PR #797, opened 2023-01-14
Add clarification text on claim conditions snapshot

nachoiacovino:
Tell me if this makes sense or you would word it differently @jnsdls


contract-ui/tabs/claim-conditions/components/snapshot-upload.tsx
```diff
@@ -341,12 +333,16 @@ export const SnapshotUpload: React.FC<SnapshotUploadProps> = ({
...
-                        </Text>
                       </>
                     )}
+                    <Text as={ListItem}>
+                      Repeated addresses will be removed and only the first
+                      found will be kept.
+                    </Text>
+                    <Text as={ListItem}>
+                      The limit you set is for the maximum amount of NFTs a
+                      wallet can claim, not how many they can receive in total.
```
jnsdls, line 344: this will show for both old and new drops I believe? It's only true for new drops.

nachoiacovino: Well, for old drops is "how many can claim _per transaction_" but the phrase still holds true.

isn't that right?

jnsdls: let's add the "per transaction" for the old one, then it's good to ship

PR #800, opened 2023-01-16
feat: add Leanne Bats as mentor

avneesh0612:


nachoiacovino:
@avneesh0612 can we make the image smaller and optimize it with tinypng?

avneesh0612:
sure

PR #814, opened 2023-01-23
Any EVM + middleware

MananTank:
## Known Issues

* Network Selection dropdown is not working for the Coinbase wallet for non-default chains

## Enhancements 

* move `CURRENCIES` object to chains package
* replace `<select>` with a custom combo box in the network switcher


components/shared/CurrencySelector.tsx
```diff
@@ -19,24 +20,38 @@ export const CurrencySelector: React.FC<CurrencySelectorProps> = ({
...
+        return !CURRENCIES[chainId as SUPPORTED_CHAIN_ID]?.find(
+          (currency: CurrencyMetadata) =>
+            currency.address.toLowerCase() === initialValue.toLowerCase(),
+        );
+      }
+
+      // QUESTION - is this correct?
+      // for non-default chains
+      return true;
     }
```
MananTank, line 41: @jnsdls Need your input here. is this correct?

constants/currencies.ts
```diff
@@ -401,6 +401,16 @@ const SolanaDevnet: CurrencyMetadata[] = [
...
+// Question: what date should here?
+// Temporary fake data for creating a build
+const FakeData: CurrencyMetadata[] = [
+  {
+    address: "LOCALHOST_TOKEN",
+    name: "localhost",
+    symbol: "ETH",
+  },
+];
+
```
MananTank, line 413: @jnsdls Need your input here. What should be the metadata for localhost and hardhat, or should they be removed from the CURRENCIES object?

constants/rpc.ts
```diff
@@ -64,6 +64,11 @@ export const EVM_RPC_URL_MAP: Record<SUPPORTED_CHAIN_ID, string> = addAPIKey({
     process.env.SSR_RPC_BINANCE_TESTNET ||
     process.env.NEXT_PUBLIC_RPC_BINANCE_TESTNET ||
     `https://binance-testnet.${RPC_ENV}.thirdweb.com`,
+  // Question: what URL should we use for Hardhat and Localhost?
+  // temporary data for creating a build
+  // it has to be an actual URL - otherwise build will fail
+  [ChainId.Hardhat]: `https://goerli.${RPC_ENV}.thirdweb.com`,
+  [ChainId.Localhost]: `https://goerli.${RPC_ENV}.thirdweb.com`,
```
MananTank, line 71: @jnsdls need your input here, what should be RPC URL for localhost and hardhat here?
or should we remove them from this object?

pages/contracts/[...catchAll].tsx
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export const getServerSideProps: GetServerSideProps = async () => {
+  return {
+    redirect: {
+      destination: "/explore",
+      permanent: false,
+    },
+  };
+};
```
Comment (MananTank) [inline pages/contracts/[...catchAll].tsx:15]: @jnsdls What do you think about this approach for redirecting old paths? 

I decided to NOT do this redirect via middleware because that would require ignoring valid routes eg: /contracts/deploy and /contracts/release - which seems like an unnecessary coupling

constants/rpc.ts
jnsdls: This object should no longer be necessary and we should instead rely on the configured chains (and our chains package includes localhost config)

pages/contracts/[...catchAll].tsx
Comment (jnsdls) [reply pages/contracts/[...catchAll].tsx:15]: I think this approach is fine

constants/currencies.ts
jnsdls: This should come from the chain config not from a constants map

components/shared/CurrencySelector.tsx
jnsdls: This seems correct, yes

constants/currencies.ts
jnsdls: With that said I think we do need to add in currencies that we know for our default chains, so really this becomes a job of merging those for chains we know special currency addresses for

jnsdls: That should probably happen still in the chains package though…

PR #823, opened 2023-01-26
Support for NFTs displaying 3d models

JustinTime42:


nachoiacovino:
We should also add support for it in batch upload (which lives in batch.ts) 

Also reminds me we should have a centralized place for this.

PR #828, opened 2023-01-30
Fix Infinite rendering issue of <BuiltinContractForm/>

MananTank:
`<BuiltinContractForm/>` is rendered one after another indefinitely because of the `useDeploy` hook


jnsdls (approved):
Makes sense, LGTM

PR #834, opened 2023-02-02
Redesign contract functions overview

shift4id:


contract-ui/tabs/explorer/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+  contractAddress,
+}) => {
+  const { contract, isSuccess, isError } = useContract(contractAddress);
+
+  const functions = useContractFunctions(contract?.abi as Abi);
+  if (!contractAddress) {
+    return <div>No contract address provided</div>;
+  }
+
+  if ((!contract?.abi && isSuccess) || isError) {
```
nachoiacovino, line 22: do we need to do this here? right now we only do it on overview

jnsdls (approved):
LGTM

contract-ui/tabs/explorer/page.tsx
jnsdls: yeah only overview is fine for this

nachoiacovino (approved):
love the new design, looks great on its own tab

nachoiacovino:
one improvement, I will make only the functions scroll, not the whole thing, would look better on the releases page 
image

PR #861, opened 2023-02-14
fix: NFT Overview empty state and hide Contract Card 

shift4id:


components/explore/contract-card/index.tsx
```diff
@@ -66,7 +66,10 @@ export const ContractCard: React.FC<ContractCardProps> = ({
     return replaceDeployerAddress(h);
   }, [contractId, publisher, version, via]);
 
-  return (
+  console.log(publishedContractResult);
```
nachoiacovino, line 69: remove this plss

PR #871, opened 2023-02-15
Fix Solana programs

nachoiacovino:


jnsdls (approved):
Makes sense, thanks nacho!

PR #880, opened 2023-02-16
feat: Add Token + Marketplace Details to Contract Overview

shift4id:


contract-ui/tabs/listings/components/listing-stats.tsx
```diff
@@ -0,0 +1,90 @@
...
+      </Card>
+    </SimpleGrid>
+  );
+};
+
+interface ListingStatsV3Props {
+  contract?: MarketplaceV3;
+}
+
+export const ListingStatsV3: React.FC<ListingStatsV3Props> = ({ contract }) => {
```
nachoiacovino, line 45: you need to detect here if the contract support these extensions

if direct listings, show count for direct listings
if english auctions, show count for english auctions
if both, show total listings count

contract-ui/tabs/overview/components/MarketplaceDetails.tsx
```diff
@@ -0,0 +1,366 @@
...
+  const { contract } = useContract(contractAddress, contractType);
+
+  if (contractType === "marketplace" && contract) {
+    return (
+      <MarketplaceV1Details
+        contract={contract as Marketplace}
+        trackingCategory={trackingCategory}
+      />
+    );
+  } else if (contractType === "marketplace-v3" && contract) {
```
nachoiacovino, line 84: never check for `contractType === "marketplace-v3"`, you need to detect if either direct listings or english auctions extension exist, then you show this.

contract-ui/tabs/overview/components/MarketplaceDetails.tsx
```diff
@@ -0,0 +1,366 @@
...
+            isLoading={listingsQuery.isLoading}
+            trackingCategory={trackingCategory}
+          />
+        </>
+      )}
+    </Flex>
+  );
+};
+
+const MarketplaceV3Details: React.FC<
```
nachoiacovino, line 153: same thing here, we need to detect the extensions

contract-ui/tabs/overview/components/MarketplaceDetails.tsx
```diff
@@ -0,0 +1,366 @@
...
+    displayValue: "0",
+  },
+  startTimeInSeconds: BigNumber.from(idx),
+});
+
+interface ListingCardsProps {
+  listings: ListingData[];
+  isLoading: boolean;
+  trackingCategory: TrackedLinkProps["category"];
+  isMarketplaceV3?: boolean;
```
nachoiacovino, line 279: this boolean should be isMarketplaceV1, that's the one we have a special case for, marketplace v3 is not a "prebuilt" per se but a combination of different contract extensions

contract-ui/tabs/overview/page.tsx
```diff
@@ -35,12 +37,24 @@ export const CustomContractOverviewPage: React.FC<
...
+
   if (!contractAddress) {
     return <div>No contract address provided</div>;
   }
+
   return (
     <SimpleGrid columns={{ base: 1, xl: 4 }} gap={8}>
       <GridItem as={Flex} colSpan={{ xl: 3 }} direction="column" gap={16}>
+        {contract &&
+          ["marketplace", "marketplace-v3"].includes(contractTypeData) && (
```
nachoiacovino, line 49: again never check against marketplace-v3

contract-ui/tabs/overview/components/MarketplaceDetails.tsx
```diff
@@ -0,0 +1,366 @@
...
+          </Flex>
+          <ListingCards
+            listings={directListings}
+            isLoading={directListingsQuery.isLoading}
+            trackingCategory={trackingCategory}
+            isMarketplaceV3
+          />
+        </>
+      )}
+      {!englishAuctionsQuery.isLoading &&
```
nachoiacovino, line 226: not sure why since I don't see anything weird at first glance with the code but english auctions wasn't loading for me on my testing

nachoiacovino (dismissed):
after fixing the problem with english auctions not rendering, this would work for all people deploying marketplace v3 contracts directly.

But we need to detect the extension because we are expecting people to not necessarily have both extensions and deploy their own versions of the contract and still be able to see the overview.

contract-ui/tabs/overview/components/MarketplaceDetails.tsx
nachoiacovino: here's the url of my marketplace to see it: https://thirdweb-www-git-sid-contract-overview-token.thirdweb-preview.com/mumbai/0xCd3561F21E43f285c0f3ae1DA08d5C0BC1254660

nachoiacovino (approved):
can we capitalize properly here?
image

Also I'm wondering if it makes sense for marketplace v3 to say what type of listing are they in the card since we have one section for each
image

I just asked Bobby about it.

Neither of these are blockers but would be nice to fix the capitalization thing since it's one second, other than that seems good to merge.

Other than that, we're default to english auctions in some places and that's gonna break as soon as we release dutch auctions, might've been better to default to undefined but it's not a big deal since we're definitely touching all these sections once dutch auctions gets added.

jnsdls:
@shift4id please fix the small issues @nachoiacovino mentioned once this is merged, don't want to delay merging :)

PR #909, opened 2023-02-23
next 13.2?

jnsdls:
https://nextjs.org/blog/next-13-2


jnsdls:
@MananTank @nachoiacovino this needs functionality tested - maybe u guys can bring it to testjam it's just dependency upgrades I'm hoping it may fix some underlying issues that are causing the sentry errors - also in general just good to stay up to date on things

PR #936, opened 2023-03-02
Update to nightly

adam-maj:


adam-maj:
We shouldn't have had to make the changes in here 🤔 We must have touched Solana somehow?

PR #959, opened 2023-03-09
Support ENS on all functions

adam-maj:


jnsdls:
_does it actually work ser?_

jnsdls:
done via nightly

PR #967, opened 2023-03-11
add initial analytics

jnsdls:


jnsdls:
will re-visit with new API

PR #973, opened 2023-03-14
Fix button overflowing on long chain names

nachoiacovino:


components/buttons/MismatchButton.tsx
```diff
@@ -228,8 +235,12 @@ const MismatchNotice: React.FC<{
         isDisabled={!actuallyCanAttemptSwitch}
         colorScheme="orange"
         textTransform="capitalize"
+        noOfLines={1}
```
jnsdls, line 238: won't this alone already do it?

jnsdls (commented):
hmm

components/buttons/MismatchButton.tsx
nachoiacovino: it gets cut out (doesn't respect paddings) so I went through the convoluted solution but still added that for mobile or edge cases)

basically it's the "this ensures it never happens no matter what"

PR #1040, opened 2023-03-28
Stop UI flashing, use react hooks for permissions

nachoiacovino:


hooks/useDelayedDisplay.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useState } from "react";
+
+const useDelayedDisplay = (delay: number) => {
```
jnsdls, line 3: this seems to be doing very similar things to our existing `<ClientOnly>` component (since it will also only ever turn true on the client) - can we instead just use that wherever we're using this?

jnsdls (approved):
shippit

PR #1046, opened 2023-03-30
Make the selector a dropdown on the sidebar on build tab

nachoiacovino:
Making it a draft because not sure how to make the sidebar sticky :D


jnsdls:
sticky not as important as making it work on mobile :)

PR #1068, opened 2023-04-04
Update call snippets for react and typescript

adam-maj:


nachoiacovino (commented):
I think you missed it for the web3button snippet?

PR #1128, opened 2023-04-19
Samu sarmiento ambassadors page

SamuSarmiento:


nachoiacovino:
hey @SamuSarmiento thank you for the contribution!

we don't use package-lock, would you be able to remove it? also if you're only doing incremental updates we can also revert back those changes so there's no changes on either package.json or the yarn.lock, then I'll approve the run. thanks!

SamuSarmiento:
> hey @SamuSarmiento thank you for the contribution!
> 
> we don't use package-lock, would you be able to remove it? also if you're only doing incremental updates we can also revert back those changes so there's no changes on either package.json or the yarn.lock, then I'll approve the run. thanks!

Most of the changes were simply to be able to develop in my environment and configuration. It should be no problem to simply add the /ambassadors index and related assets. The first Commit.

pages/ambassadors/index.tsx
```diff
@@ -0,0 +1,573 @@
...
+          </Flex>
+        </HomepageSection>
+        <NewsletterSection />
+        <HomepageFooter />
+      </Flex>
+    </DarkMode>
+  );
+};
+
+Ambassadors.pageId = PageId.Learn;
```
SamuSarmiento, line 570: Ambassadors.pageId = PageId.Ambassadors

SamuSarmiento: @nachoiacovino If you merge just this /ambassadors/index.tsx it can work? With the related assets

nachoiacovino: I can't nitpick your PR unless I literally copy the changes and push them myself which will take your commits away, do you think you could fix the things I said so I can deploy and check if there's any more changes needed?

PR #1189, opened 2023-05-09
Add support for ZKSync Era Testnet

kumaryash90:


components/contract-components/hooks.ts
```diff
@@ -531,13 +536,31 @@ export function useCustomContractDeployMutation(
...
-            forceDirectDeploy,
-          },
-        );
+        // Handle ZkSync deployments separately
+        const isZkSync =
+          chainId === ZksyncEraTestnet.chainId || chainId === ZksyncEra.chainId;
+        if (isZkSync) {
+          // Get metamask signer using zksync-web3 library -- for custom fields in signature
+          const zkSigner = new Web3Provider(
+            window.ethereum as unknown as providers.ExternalProvider,
```
MananTank, line 545: This assumes that user is using injected wallet, but that is not always true right?

PR #1192, opened 2023-05-10
Filter dashboard by network

mykcryptodev:
My dashboard is getting crowded with lots of deployments and with Goerli being deprecated soon, I have a feeling that I'll be scrolling through a lot of deprecated contracts.

I created this filter-by-network feature and I'm happy to get feedback on it :) 

https://www.loom.com/share/012ed260351645c781326cf9553a9b9e


jnsdls:
idea is right but the filter should be part of the table itself, see: https://tanstack.com/table/v8/docs/examples/react/filters for an example

mykcryptodev:
> idea is right but the filter should be part of the table itself, see: https://tanstack.com/table/v8/docs/examples/react/filters for an example

@jnsdls I updated the filter to work directly on the table. The filter now replaces the header in the table but I'm happy to take direction on the design if that doesn't vibe. I ended up implementing the table's native pagination and removed the code that  simulated pagination. I did this in order for the filter to work across the entire dataset even if that data is not yet on the current page.

Let me know what ya think :)

MananTank:
@mykcryptodev - Can you add proper types to fix the build error?

jnsdls:
@mykcryptodev here's the error: 
image

mykcryptodev:
> @mykcryptodev here's the error: image

@jnsdls I fixed the type errors for the build

components/contract-components/tables/deployed-contracts.tsx
```diff
@@ -86,10 +86,26 @@ export const DeployedContracts: React.FC<DeployedContractsProps> = ({
   contractListQuery,
   limit = 10,
 }) => {
+  const [showMoreLimit, setShowMoreLimit] = useState(limit);
```
mykcryptodev, line 89: @MananTank I like what you did with the design!

I think this commit introduced a regression bug though. I changed around how Show More works because the filter needs to be able to filter all of your contracts (even the ones that are not on the current page). Going back to this method of Show More will only filter contracts on the current page. Is there a good reason to revert back to this method of Show More?

MananTank: @mykcryptodev I've taken care of it - `chainIdsWithDeployments` is calculated with `contractListQuery.data` and not the sliced data

MananTank:
LGTM, Merging it
Thanks @mykcryptodev for your contribution 🔥

components/contract-components/tables/deployed-contracts.tsx
MananTank: my bad - you are right. we can't have show more the way list is filtered

mykcryptodev: @MananTank Im on it - I can open a PR -- logged issue #1204

MananTank: I pushed a fix - https://github.com/thirdweb-dev/dashboard/pull/1205

PR #1304, opened 2023-06-15
Replace any-evm image with optimized version

MananTank:


jnsdls:
looks a bit pixel-y but lets ship for now

PR #1306, opened 2023-06-15
Add Progress Bar

MananTank:


MananTank:
Need to fix a few things here - not ready yet

jnsdls:
Feels a bit thick on mobile, can we slim it down?

Maybe can we see if the transition is fast and then not show it, only show it after a short delay? (For long transitions)

components/shared/ProgressBar.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const [progress, setProgress] = useState(0);
+  const [isVisible, setIsVisible] = useState(false);
+
+  // Progress bar inline styling
+  const styling = {
+    position: "fixed",
+    top: 0,
+    left: 0,
+    width: `${progress}%`,
+    height: "3px",
```
jnsdls, line 31: let's try 2px here

components/shared/ProgressBar.tsx
```diff
@@ -0,0 +1,104 @@
...
+  } as React.CSSProperties;
+  useEffect(() => {
+    // Declare timeout
+    let status: "in-progress" | "idle" = "idle";
+    let intervalId: NodeJS.Timeout;
+
+    // Route change start function
+    const onRouteChangeStart = () => {
+      status = "in-progress";
+      setIsVisible(true);
```
jnsdls, line 45: can we delay this by let's say 100ms or 200ms so that if the route change completes before we don't show it?

jnsdls (commented):
I think overall makes sense I left 2 small notes

jnsdls (approved):
LGTM :shipit:

PR #1309, opened 2023-06-16
Default trusted forwarders for zksync

kumaryash90:


joaquim-verges (commented):
Does OZ defender support zksync? Wonder if there's any relayer that works rn 😅 also with their AA setup, it might work a bit differently no?

Did we have a report of this being an issue already?

kumaryash90:
@joaquim-verges yeah they had enabled zksync some time ago: https://forum.openzeppelin.com/t/defender-release-support-for-zksync-era/35565

Though we haven't yet received any reports related to gasless tx issues, but adding this just in case.
Will look at ZKSync AA, and if gasless is needed at all.

joaquim-verges:
Nice, good catch anticipating this. Ship it

components/contract-components/hooks.ts
```diff
@@ -543,24 +544,30 @@ export function useCustomContractDeployMutation(
...
+                chainId,
+                compilerMetadata.data?.name,
+              )
+            : await getTrustedForwarders(
+                sdk.getProvider(),
+                sdk.storage,
+                compilerMetadata.data?.name,
+              );
+
+          console.log("trusted forwarders: ", trustedForwarders);
```
nachoiacovino, line 565:
```suggestion
```

PR #1348, opened 2023-06-29
Rename smart wallets of the SDK -> accounts

nkrishang:


public/robots.txt
```diff
@@ -2,10 +2,6 @@
 User-agent: *
 Allow: /
 
-# *
```
nachoiacovino, line 5: curious why this got changed?

nachoiacovino (approved):
everything looks good except the change in robots.txt but I've seen that happening before when building, can you revert that before merging?

public/robots.txt
nkrishang: not sure why this change, let me check

PR #1354, opened 2023-07-01
Update README.md

kienngo98:
Trying to run this command on Windows and get the following error (image)
image

This can be solved with the `--ignore-engines` flag. This issue is somewhat similar with https://github.com/thirdweb-dev/js/issues/1179


kienngo98:
Might be a better idea to just update node version instead of ignoring-enginges

PR #1376, opened 2023-07-06
Improve /chainlist perf

MananTank:
Before:

https://www.webpagetest.org/breakdownTimeline.php?test=230706_AiDcJP_7M3&run=1

image

After: 

https://www.webpagetest.org/breakdownTimeline.php?test=230706_BiDcV6_7E3&run=3

image


MananTank:
Testing...

PR #1430, opened 2023-07-20
Pass Chain icon to embed via query param

MananTank:


contract-ui/tabs/embed/components/embed-setup.tsx
```diff
@@ -30,6 +30,21 @@ import {
   Heading,
 } from "tw-components";
 
+export function minimizeChain(chain: Chain) {
```
jnsdls, line 33: let's instead change this in the package please

.env.example
```diff
@@ -6,6 +6,9 @@ NEXT_PUBLIC_THIRDWEB_API_HOST="http://localhost:3005"
 # staging
 NEXT_PUBLIC_THIRDWEB_API_HOST="https://api.staging.thirdweb.com"
 
+# Get the client id from https://thirdweb.com/create-api-key and set it here, Make sure to allow localhost:3000 in the allowed origins
+NEXT_PUBLIC_DASHBOARD_CLIENT_ID=
```
jnsdls, line 10: great!

jnsdls (changes requested):
let's move the change to the chains package

contract-ui/tabs/embed/components/embed-setup.tsx
MananTank: Moved it to dashboard because that function is only used in dashboard

MananTank: Done

jnsdls (approved):
LGTM

PR #1432, opened 2023-07-21
chore: sentry cleanup

assimovt:


jnsdls:
#1435 should have taken care of most of this :)

PR #1469, opened 2023-08-01
Fix OG image not showing contract with name <=3 characters

MananTank:
Fixes https://linear.app/thirdweb/issue/FTD-1950/og-image-for-contract-doesnt-pull-contract-name-correctly


jnsdls (approved):
LGTM

PR #1556, opened 2023-08-17
Replace multiwrap with loyalty-card on explore

nkrishang:


data/explore.ts
```diff
@@ -33,7 +33,7 @@ const NFTS = {
   description:
     "NFT Collections, Editions, Drops and everything else NFT-related.",
   contracts: [
-    "thirdweb.eth/Multiwrap",
+    "thirdweb.eth/LoyaltyCard",
```
joaquim-verges, line 36: can you still leave multiwrap at the end of the list plz

PR #1559, opened 2023-08-17
feat: billing

assimovt:


components/settings/Account/config.ts
```diff
@@ -0,0 +1,4 @@
+export const Settings = {
+  CustomerPortalUrl:
+    "https://billing.stripe.com/p/login/test_8wMeWi3w9fTd7D2288",
```
assimovt, line 3: FIXME: Needs to come from .env

components/settings/Account/Usage.tsx
```diff
@@ -0,0 +1,63 @@
...
+}) => {
+  const bundlerTotal = useMemo(() => {
+    if (!usage?.bundler || usage.bundler.length === 0) {
+      return <i>No usage in this period</i>;
+    }
+    const total = usage.bundler
+      .map((u) => u.sumTransactionFeeUsd)
+      .reduce((acc, num) => acc + num, 0);
+
+    return numeral(total).format("($ 0.00 a)");
```
arcoraven, line 28: Rather than installing a library, can we just [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)?

```typescript
console.log(new Intl.NumberFormat(undefined, { style: "currency", currency: "USD" }).format("12345.67"));
// $12,345.67 in my locale
```

components/dashboard/StepsCard.tsx
```diff
@@ -39,13 +41,16 @@ export const StepsCard: React.FC<StepsCardProps> = ({
...
         flexDir="column"
-        py={6}
-        px={6}
+        p={8}
         gap={4}
         position="relative"
         cursor={isComplete ? (isOpen ? "default" : "pointer") : undefined}
         onClick={isComplete ? () => !isOpen && setIsOpen(true) : undefined}
-        bg="linear-gradient(158.84deg, rgba(255, 255, 255, 0.05) 13.95%, rgba(255, 255, 255, 0) 38.68%)"
+        bg={
```
Marfuen, line 49: @assimovt You can actually use useColorModeValue, to get the value. Just nicer syntax.

```
  const bg = useColorModeValue('red.500', 'red.200')
  // Then use the value returned, in this example red.500 is light mode and red.200 is dark mode
```

components/settings/Account/Usage.tsx
```diff
@@ -0,0 +1,102 @@
...
+  usage: UsageByService | undefined;
+  usageLoading: boolean;
+}
+
+export const Usage: React.FC<UsageProps> = ({ usage, usageLoading }) => {
+  const bundlerTotal = useMemo(() => {
+    if (!usage?.bundler || usage.bundler.length === 0) {
+      return <i>N/A</i>;
+    }
+    const total = usage.bundler?.[0].billableUsd || 0;
```
Marfuen, line 17: Why are we grabbing the first element here?

components/settings/Account/Usage.tsx
```diff
@@ -0,0 +1,102 @@
...
+      currency: "USD",
+    }).format(total);
+  }, [usage]);
+
+  return (
+    <VStack gap={8} w="full">
+      {usageLoading && <Spinner size="sm" />}
+      {!usageLoading && (
+        <VStack w="full" gap={12}>
+          {/* <VStack alignItems="flex-start" gap={6} w="full">
```
Marfuen, line 30: Let's remove commented code if we don't need it

components/settings/Account/Usage.tsx
```diff
@@ -0,0 +1,102 @@
...
+              <UsageCard
+                title="Smart Wallets"
+                description="Total USD consumed"
+                total={bundlerTotal}
+                tooltip="Explain here how we calculate and charge for Bundler?"
+              />
+            </SimpleGrid>
+          </VStack>
+
+          {/* <VStack alignItems="flex-start" gap={6} w="full">
```
Marfuen, line 80: Same here, what's the purpose of keeping it?

components/settings/Account/Usage.tsx
assimovt: yeah long story, but the format of usage is weirdly delivered from API now, per chain, but total for each chain. we need to fix API structure when we get to other services.

components/settings/Account/Usage.tsx
assimovt: its coming soon, this week

components/settings/Account/Usage.tsx
assimovt: its coming soon, this week

PR #1580, opened 2023-08-21
Add analytics to overview tab

adam-maj:


@3rdweb-sdk/react/components/popularChains.ts
```diff
@@ -5,6 +5,7 @@ import {
   AvalancheFuji,
   BaseGoerli,
   Binance,
+  Base,
```
nachoiacovino, line 8: you're importing here but then not passing it to the list of popular chains?

data/analytics/hooks.ts
```diff
@@ -10,6 +10,12 @@ export type AnalyticsQueryParams = {
   interval?: "minute" | "hour" | "day" | "week" | "month";
 };
 
+// TODO: Keep updated with actual ClickHouse data
+export const SUPPORTED_ANALYTICS_CHAINS = [
```
nachoiacovino, line 14: would be nice to do Ethereum.ChainId etc instead of the actual number for better redeability

nachoiacovino (commented):
would put under the nfts/listings tab instead of as the first thing

PR #1595, opened 2023-08-28
Group functions by extensions in contract explorer

joaquim-verges:


components/contract-functions/contract-function.tsx
```diff
@@ -144,27 +152,78 @@ interface ContractFunctionsPanelProps {
...
-  const viewFunctions: AbiFunction[] = useMemo(() => {
-    return fnsOrEvents.filter(
-      (fn) =>
-        (fn as AbiFunction).stateMutability === "pure" ||
-        (fn as AbiFunction).stateMutability === "view",
-    ) as AbiFunction[];
-  }, [fnsOrEvents]);
+  const writeFunctions: ExtensionFunctions[] = useMemo(() => {
+    return functionsWithExtension
+      .map((e) => {
```
assimovt, line 193: nit: reduce would be simpler here, so no need to filter out undefines

components/contract-functions/contract-function.tsx
```diff
@@ -144,27 +152,78 @@ interface ContractFunctionsPanelProps {
...
+        if (filteredFunctions.length === 0) {
+          return undefined;
+        }
+        return {
+          extension: e.extension,
+          functions: filteredFunctions,
+        };
+      })
+      .filter((e) => e !== undefined) as ExtensionFunctions[];
+  }, [functionsWithExtension]);
```
assimovt, line 208: nit: line breaks between functions. will add to lint rules

assimovt (approved):
lgtm

components/contract-functions/contract-function.tsx
joaquim-verges: prettier did this though

PR #1597, opened 2023-08-28
chore: don't dismiss drawer on outside click

assimovt:
fixes FTD-2072


joaquim-verges (approved):
looks good @assimovt 🚢

PR #1635, opened 2023-09-08
Use sdk functions to resolve ens

nachoiacovino:


lib/ens.ts
```diff
@@ -32,21 +16,45 @@ export async function resolveAddressToEnsName(
 ): Promise<ENSResolveResult> {
   invariant(utils.isAddress(address), "address must be a valid address");
 
-  return {
-    ensName: await getMainnetProvider().lookupAddress(address),
-    address,
-  };
+  try {
+    const ensName = await resolveAddress(address);
```
jnsdls, line 20: which clientId / secret key does this use?

lib/ens.ts
```diff
@@ -32,21 +16,45 @@ export async function resolveAddressToEnsName(
...
   ensName: string,
 ): Promise<ENSResolveResult> {
   invariant(isEnsName(ensName), "ensName must be a valid ens name");
 
-  return {
-    ensName,
-    address: await getMainnetProvider().resolveName(ensName),
-  };
+  try {
+    const address = await resolveEnsSdk(ensName);
```
jnsdls, line 43: which clientId / secret key does this use?

nachoiacovino: whatever we use for the rest of the dashboard, anyway we can't do this right now because resolveEns in the sdk is fake and doesn't actually resolveEns, both resolveEns and resolveAddress returns the address from the ens 🙃

PR #1642, opened 2023-09-11
Fix light theme condition

kien-ngo:


MananTank:
LGTM, thanks @kien-ngo

PR #1700, opened 2023-09-25
fix: Update Broken Link

Shiva-Sai-ssb:
I've resolved the broken link for the NFT Gallery Demo website by replacing it with the new link from the GitHub repository.


jnsdls:
thanks for the contribution @Shiva-Sai-ssb - this is actually a misconfiguration on the example, it *should* be `.thirdweb-example.com` I've fixed that now. :)

PR #1747, opened 2023-10-05
[DO NOT MERGE] Testing api-server

gerrysaporito:
## Changes

- Added extra fetch call to `useConfirmEmail()` which now calls the paper sign-in endpoint _after_ validating OTP

## Testing

- [x] Sign in with valid OTP - accept OTP and close modal (as expected)
- [x] Sign in with invalid OTP - unsuccessful OTP and keep modal opened (as expected)


@3rdweb-sdk/react/hooks/useApi.ts
```diff
@@ -302,6 +305,19 @@ export function useConfirmEmail() {
         throw new Error(json.message);
       }
 
+      // Add account with validated email to the Paper database as a seller
```
jnsdls, line 308: I wonder if we could do this from server -> server?

is there something that requires this to happen from the client?

gerrysaporito: Moved to api-server 👍

PR #1781, opened 2023-10-13
AWS logo on website

juandolealt:


juandolealt:
once this goes through will make another PR to add it to the logo carousel

PR #1804, opened 2023-10-18
Add trailing slash to IPFS URI for updateBatchBaseURI

kien-ngo:
Problem:
If the user enters the URI: `ipfs://Qmc...End` (without the trailing slash), everything will be broken because the new tokenURI will become:
- `ipfs://Qmc...End0` ("0" is appended at the end without "/")
- `ipfs://Qmc...End1` ("1" is appended at the end without "/")

This small change is to ensure that the trailing slash is automatically added at the end of the string.

Fwiw, if the user enters an invalid folder URI, things will be broken anyway


kien-ngo:
@jnsdls I think I need to be invited to Vercel to pass this step: `Vercel — Deployment has failed`?

PR #1826, opened 2023-10-24
refactor(chains): use the api to fetch chains instead of relying on chains package

jnsdls:


jnsdls:
requires API server changes to roll out first

PR #1855, opened 2023-10-26
Don't render ghost tooltip

nachoiacovino:


components/engine/overview/transactions-table.tsx
```diff
@@ -148,15 +151,17 @@ export const TransactionsTable: React.FC<TransactionsTableProps> = ({
...
-                <Card bgColor="backgroundHighlight">
-                  <Text>
-                    {status === "errored"
-                      ? errorMessage
-                      : status === "mined" && minedAt
-                      ? `Completed ${format(new Date(minedAt), "PP pp")}`
-                      : undefined}
-                  </Text>
-                </Card>
+                shouldShowTooltip ? (
```
jnsdls, line 154: better would be to not render the tooltip component at all in this case, so you could conditionally wrap tag in tooltip or not

PR #1904, opened 2023-11-04
fix code snippets url resolution

lakgarg:


contract-ui/components/build-apps-button.tsx
```diff
@@ -12,11 +12,17 @@ export const BuildAppsButton: ComponentWithChildren<BuildAppsButtonProps> = ({
   ...restButtonProps
 }) => {
   const { query, asPath } = useRouter();
-  const path = useMemo(() => {
-    const [network, address] = query.paths as string[];
 
-    return `/${network}/${address}/code`;
-  }, [query.paths]);
+  const [path] = useMemo(() => {
```
nachoiacovino, line 16: thank you for fixing this! we can simpify it though and also make it consistent

let's get the network with useDashboardEVMChainId and pass the contractAddress from the parent component? what do you think?

PR #1979, opened 2023-11-17
refactor: simplify middleware route matching

jnsdls:
- we have a narrow set of statically know-able paths that we need to run through middleware
- this changes the middleware matcher from *broadly* matching on every request to instead matching only those set of paths


MananTank (approved):
LGTM

PR #2019, opened 2023-11-28
Isaac/add payment webhooks

IDubuque:
- add payments to webhooks


components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +15,76 @@ interface PaymentsWebhooksProps {
...
   accountId,
 }) => {
-  return <Flex flexDir="column" gap={8}></Flex>;
+
+  const [productionWebhooks, setProductionWebhooks] = React.useState<PaymentsWebhooksType[]>([]);
+  const [testnetWebhooks, setTestnetWebhooks] = React.useState<PaymentsWebhooksType[]>([]);
+
+  const { data: webhooks, isLoading, isFetched, error } = usePaymentsWebhooksByAccountId(accountId);
+
+  React.useEffect(() => {
```
nachoiacovino, line 119: never use useEffect and states for things that can be got from a different place, you can use a useMemo here which is much better

`const { productionWebhooks, testnetWebhooks } = useMemo(() => ... return { productionWebhooks, testnetWebooks}, [webhooks])`

lmk if that makes sense

components/payments/settings/payment-webhooks.tsx
```diff
@@ -1,4 +1,12 @@
-import { Flex } from "@chakra-ui/react";
+import { usePaymentsWebhooksByAccountId } from "@3rdweb-sdk/react/hooks/usePayments";
+import type { PaymentsWebhooksType } from "@3rdweb-sdk/react/hooks/usePayments";
+import { Flex, Divider } from "@chakra-ui/react";
+import {
+  Card,
+  Heading
+} from "tw-components";
+import { PaymentsWebhooksTable, PaymentsWebhooksTableProps } from "./payments-webhooks-table";
+import React from "react";
```
nachoiacovino, line 17: no need to import React, just import what you need

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +15,76 @@ interface PaymentsWebhooksProps {
...
+        <Heading>Webhooks</Heading>
+      </Flex>
+
+      <Divider />
+
+      <Flex flexDir="column" gap={2}>
+        <Heading size={"title.md"}>Production</Heading>
+        <PaymentsWebhooksTable
+          webhooks={productionWebhooks}
+          onCreate={createWebhookHandlerFactory(true)}
```
nachoiacovino: this would just automatically get called on render, that seems like a mistake?

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -211,15 +235,56 @@ export const PaymentsWebhooksTable: React.FC<PaymentsWebhooksTableProps> = ({
...
-      isFetched={isFetched}
-    />
+    <>
+      <Modal isOpen={isOpen} onClose={onClose} isCentered>
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader>Delete webhook</ModalHeader>
+          <ModalCloseButton />
+          <ModalBody>
+            {webhookToRevoke && (
```
nachoiacovino, line 244: what if you... don't have a webhook to revoke? 

I don't think this is the way to go about this, can you check how I did the delete confirmation in engine and do something similarr? You can check the RemoveModal on `admins-table` you should have the ID before clicking the button so you can always render the modal knowing what you're deleting

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -78,6 +90,9 @@ export const PaymentsWebhooksTable: React.FC<PaymentsWebhooksTableProps> = ({
   const editWebhookRef = React.useRef<HTMLInputElement>(null);
   const createWebhookRef = React.useRef<HTMLInputElement>(null);
 
+  const [webhookToRevoke, setWebhookToRevoke] = React.useState<TableWebhookType>();
+  const { isOpen, onOpen, onClose } = useDisclosure();
+
   React.useEffect(() => {
```
nachoiacovino: same for this, we shouldn't need to keep any state of what we're editing do we? can't we just pass it in to the component when opening the modal?

components/payments/settings/payment-webhooks.tsx
```diff
@@ -30,6 +33,12 @@ export const PaymentsWebhooks: React.FC<PaymentsWebhooksProps> = ({
   const { mutate: updateWebhook, isLoading: isUpdateLoading } = usePaymentsUpdateWebhook(accountId);
   const { mutate: createWebhook, isLoading: isCreateLoading } = usePaymentsCreateWebhook(accountId);
 
+  const [secretKey, setSecretKey] = React.useState<string>("");
+  React.useEffect(() => {
```
nachoiacovino, line 114: why do you need this in state at all? can you just... define it once? it will get re-computed at every render and if don't' want that to happen just put it outside the component iself (like on line 15)

components/payments/settings/payment-webhooks.tsx
```diff
@@ -29,16 +38,83 @@ export const PaymentsWebhooks: React.FC<PaymentsWebhooksProps> = ({
...
+  const { mutate: createWebhook, isLoading: isCreateLoading, error: createError, isSuccess: isCreateSuccess } = usePaymentsCreateWebhook(accountId);
+
+  const [isUpdateDelete, setIsUpdateDelete] = React.useState(false);
+  const [currentDeleteWebhook, setCurrentDeleteWebhook] = React.useState<PaymentsWebhooksType>();
+  const [currentUpdateWebhook, setCurrentUpdateWebhook] = React.useState<PaymentsWebhooksType>();
+
+  const [currentCreateWebhook, setCurrentCreateWebhook] = React.useState<PaymentsWebhooksType>();
+
+  const [showAlert, setShowAlert] = React.useState(false);
+  const [alertHeader, setAlertHeader] = React.useState("");
```
nachoiacovino, line 51: I believe you should need maximum one state here, you can write a record of error to header/description/status and then pick up the corresponding depending on the error gotten from the response call (check how I did it in KycStatus.tsx)

Also, you shouldn't really need useEffect, you should be able to use useMemo again here since the errors are already coming from the hooks.

In general we should depend on useEffect and useState veeeery rarely, having react query usually solves all our needs for it.

nachoiacovino (changes requested):
I would make a new PR cherry-picking your changes but I believe it would be good to give it a pass before getting this live.

Will have more feedback once this gets built and it can be tested! thanks Isaac, lmk if you need any help

components/payments/settings/payment-webhooks.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const setAlertTimeout = () => {
+    setTimeout(() => {
+      setShowAlert(false);
+      setAlertHeader("");
+      setAlertDescription("");
+      setAlertStatus(undefined);
+    }, 3000);
+  }
+
+  const triggerAlert = (status: "error"|"success", header: string, description: string) => {
```
nachoiacovino, line 64: also realized for what you're doing here you probably want to use a toast instead?

components/payments/settings/payment-webhooks.tsx
IDubuque: this is a factory function. Returns another function

components/payments/settings/payment-webhooks.tsx
IDubuque: Just a filler key for now

nachoiacovino:
Superseded by #2022

PR #2022, opened 2023-11-29
Isaac/payment webhooks

IDubuque:


@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -269,6 +289,13 @@ function usePaymentsApi() {
       };
     }
 
+    if (options?.isSellerApiKey) {
+      return json as {
+        data: { data: any; decrypted_key: string };
```
nachoiacovino, line 294: any way we can get the data type object? either hardcode or better if we can get it from the generated graphql

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+  sellerId: string;
+  url: string;
+  isProduction: boolean;
+  createdAt: Date;
+};
+
+export function usePaymentsWebhooksByAccountId(accountId: string) {
+  invariant(accountId, "accountId is required");
+
+  const address = useAddress();
```
nachoiacovino, line 904: why do we need this? we don't use address for anything in the hook

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+          sellerId: webhook.seller_id,
+          url: webhook.url,
+          isProduction: webhook.is_production,
+          createdAt: new Date(webhook.created_at),
+        })) as PaymentsWebhooksType[];
+      } else {
+        return [] as PaymentsWebhooksType[];
+      }
+    },
+    { enabled: !!paymentsSellerId && !!address },
```
nachoiacovino, line 934:
```suggestion
    { enabled: !!paymentsSellerId && !!accountId },
```

We should refetch on accountId change so we don't even have to get the address

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+        variables: {
+          sellerId: paymentsSellerId,
+        } as WebhooksBySellerIdQueryVariables,
+      });
+
+      if (error) {
+        console.error(error);
+      }
+
+      if (data && data.webhook.length > 0) {
```
nachoiacovino, line 922: can be improved a bit if you do a ternary like I do in other places in the component, but that's a nit

`return data && data?.webhook.length > 0 ? data.webhook.map(... : [] as PaymentsWebhookType[]`

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+        return data.webhook.map((webhook) => ({
+          id: webhook.id,
+          sellerId: webhook.seller_id,
+          url: webhook.url,
+          isProduction: webhook.is_production,
+          createdAt: new Date(webhook.created_at),
+        })) as PaymentsWebhooksType[];
```
nachoiacovino, line 929: the as for the data itself shouldn't be here, you need to give it in the part you did `data: any` before so it gets inherited

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+
+export type CreateWebhookInput = {
+  url: string;
+  isProduction: boolean;
+};
+
+export function usePaymentsCreateWebhook(accountId: string) {
+  invariant(accountId, "accountId is required");
+  const queryClient = useQueryClient();
+  const address = useAddress();
```
nachoiacovino, line 964: again we don't need address here

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+  const { paymentsSellerId } = useApiAuthToken();
+
+  const [createWebhookBySellerId] = useInsertWebhookMutation({
+    refetchQueries: [WebhooksBySellerIdDocument],
+  });
+
+  return useMutationWithInvalidate(
+    async (input: CreateWebhookInput) => {
+      invariant(address, "No wallet address found");
```
nachoiacovino, line 973: no need

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+      invariant(paymentsSellerId, "No seller id found");
+      invariant(isValidWebhookUrl(input.url), "Invalid webhook url");
+
+      const webhookInput: Webhook_Insert_Input = {
```
nachoiacovino, line 977: nit: if you put the object directly inside the return you get the type safety anyway and you import one less thing, but this also works

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+export type UpdateWebhookInput = {
+  webhookId: string;
+  url?: string;
+  deletedAt?: Date;
+};
+
+export function usePaymentsUpdateWebhook(accountId: string) {
+  invariant(accountId, "accountId is required");
+  const queryClient = useQueryClient();
+  const address = useAddress();
```
nachoiacovino, line 1006: same thing for the address

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+  const { paymentsSellerId } = useApiAuthToken();
+
+  const [updateWebhookBySellerId] = useUpdateWebhookMutation({
+    refetchQueries: [WebhooksBySellerIdDocument],
+  });
+
+  return useMutationWithInvalidate(
+    async (input: UpdateWebhookInput) => {
+      invariant(address, "No wallet address found");
```
nachoiacovino, line 1015: same

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+      invariant(paymentsSellerId, "No seller id found");
+
+      return updateWebhookBySellerId({
+        variables: {
+          id: input.webhookId,
+          webhookValue: {
```
nachoiacovino, line 1021: you did it here directly nice

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+            url: input.url,
+            deleted_at: input.deletedAt,
+          } as Webhook_Set_Input,
```
nachoiacovino, line 1024: but in this case I don't think you need this? you should get the type safety already

@3rdweb-sdk/react/hooks/usePayments.ts
```diff
@@ -862,3 +889,181 @@ export function usePaymentsSellerByAccountId(accountId: string) {
...
+        "api-key/get-decrypted-key",
+        {
+          sellerId: paymentsSellerId,
+        },
+        {
+          isSellerApiKey: true,
+        },
+      );
+    },
+    { enabled: !!paymentsSellerId && !!address },
```
nachoiacovino, line 1067:
```suggestion
    { enabled: !!paymentsSellerId && !!accountId },
```

and remove useAddress

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+      variant: "solid",
+      title: header,
+      description,
+      status,
+      duration: 9000,
+      isClosable: true,
+    });
+  };
+
+  const { productionWebhooks, testnetWebhooks } = useMemo(() => {
```
nachoiacovino, line 54: this is a great improvement from before!

You can only return once if you have an optional param, this looks a bit cleaner to me but both work

`
  const { productionWebhooks, testnetWebhooks } = useMemo(
    () => ({
      productionWebhooks:
        webhooks?.filter((webhook) => webhook.isProduction) || [],
      testnetWebhooks:
        webhooks?.filter((webhook) => !webhook.isProduction) || [],
    }),
    [webhooks],
  );
`

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+      return { productionWebhooks, testnetWebhooks };
+    }
+    return { productionWebhooks: [], testnetWebhooks: [] };
+  }, [webhooks]);
+
+  const onUpdateWebhook: PaymentsWebhooksTableProps["onUpdate"] = (
+    webhook,
+    newUrl,
+  ) => {
+    if (!isValidWebhookUrl(newUrl)) {
```
nachoiacovino, line 72: can't we do this validation in the form itself? I'm doing something similar for successCallbackUrl inn create-update--checkout-button.tsx. You could copy the component as it or abtract it, (line 702-731)

It's always better if we can give feedback to the user _before_ they click a button to do something

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+        `${newUrl} is not a valid webhook url, please try a different url`,
+      );
+      return;
+    }
+
+    // send the request
+    updateWebhook(
+      { webhookId: webhook.id, url: newUrl },
+      {
+        onSuccess: () => {
```
nachoiacovino, line 85: on all onSuccess and onError, you should use the useTxNotifications hook we use everywhere else, with this you don't even have to define the toast yourself, so you only needs toasts for success and error, and the other ones you have for validation, you do in the form itself (check how we it in any other component we use react query)

Also, you need to import useTrack and track every user action, something like this in this case:
`              trackEvent({
                category: "payments",
                action: "update-webhook",
                label: "success",
              });`

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+          triggerAlert(
+            "success",
+            "Webhook Created",
+            `Successfully created  ${
+              webhook.isProduction ? "production" : "testnet"
+            } webhook: ${webhook.url}`,
+          );
+        },
+        onError: () => {
+          triggerAlert(
```
nachoiacovino, line 95: add the onError from useTxNotifications and tracking:

`              trackEvent({
                category: "payments",
                action: "delete-webhook",
                label: "error",
                error,
              });`

nachoiacovino: I also feel you could every easily merge delete and update if you wanted but that's something we can look over tomorrow, it's a nice code optimization

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+        },
+      },
+    );
+  };
+
+  const createWebhookHandlerFactory = (isProduction: boolean) => {
+    const onAddWebhook: PaymentsWebhooksTableProps["onCreate"] = async (
+      url,
+    ) => {
+      if (!isValidWebhookUrl(url)) {
```
nachoiacovino, line 138: same thing to do the validation in the form itself

components/payments/settings/payment-webhooks.tsx
```diff
@@ -7,5 +21,210 @@ interface PaymentsWebhooksProps {
...
+            ) : (
+              <CodeBlock code={webhookApiKey?.data?.decrypted_key ?? ""} />
+            )
+          }
+        />
+
+        <Divider />
+
+        <Flex flexDir="column" gap={2}>
+          <Heading size={"title.md"}>Production</Heading>
```
nachoiacovino, line 204: nit, always nice if you only have a string to not use brackets
```suggestion
          <Heading size="title.md">Production</Heading>
```

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -0,0 +1,311 @@
...
+);
+
+interface EditTableRowProps {
+  webhook: PaymentsWebhooksType;
+  onEdit: (webhook: PaymentsWebhooksType, newUrl: string) => void;
+  onDelete: (webhook: PaymentsWebhooksType) => void;
+}
+
+const EditTableRow = ({ webhook, onEdit, onDelete }: EditTableRowProps) => {
+  const [editUrl, setEditUrl] = useState(webhook.url ?? "");
```
nachoiacovino, line 88: I still wonder if we really need state here for this but we can look over it tomorrow, we might do but we might find a cleaner way

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -0,0 +1,311 @@
...
+    </Tr>
+  );
+};
+
+interface CreateTableRowProps {
+  onCreate: (webhookUrl: string) => void;
+}
+
+const CreateTableRow = ({ onCreate }: CreateTableRowProps) => {
+  const [createUrl, setCreateUrl] = useState("");
```
nachoiacovino, line 170: same for this

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -0,0 +1,311 @@
...
+    onCreate(createUrl);
+    setCreateUrl("");
+  };
+
+  return (
+    <Tr borderBottomWidth={1} _last={{ borderBottomWidth: 0 }}>
+      <Td borderBottomWidth="inherit" borderBottomColor="accent.100">
+        <UrlInputField
+          webhookUrl={createUrl}
+          onChange={setCreateUrl}
```
nachoiacovino, line 183: see here would makes more sense to use useForm even if it's a simple thing (and then we don't need useState) and we let useForm handle all the logic and validation

components/payments/settings/payments-webhooks-table.tsx
```diff
@@ -0,0 +1,311 @@
...
+
+export const PaymentsWebhooksTable: React.FC<PaymentsWebhooksTableProps> = ({
+  webhooks,
+  onUpdate,
+  onDelete,
+  onCreate,
+  isLoading,
+  isFetched,
+}) => {
+  const [webhookToRevoke, setWebhookToRevoke] =
```
nachoiacovino, line 204: same thing here happy to look at this tomorrow more in depth

nachoiacovino (changes requested):
More important thing: fixing the build

All in all, amazing improvement compared to yesterday, if you start seeing me ask for nit stuff, it means I'm starting to not have enough things to complain about :D

let's pair on this tomorrow to bring it over the finish line!

components/payments/settings/payments-webhooks-table.tsx
IDubuque: I changed this from a ref because renders were clearing all the input fields

nachoiacovino (approved):
let's merge so we can test and let's improve it tomorrow with the things we went over in the call!

PR #2165, opened 2024-02-01
add frames to published contracts

easyrun42:


pages/api/frame/redirect.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  const body = (await req.json()) as RequestBody;
+
+  const metadata = untrustedMetaData.parse(body.untrustedData);
+
+  const trustedMessageByte = z.string().parse(body.trustedData?.messageBytes);
+
+  // This will throw an exception if neynar's API doesn't validate the message
+  await Warpcast.validateMessageWithReturnedFrameUrl(trustedMessageByte);
```
joaquim-verges: do we actually need to do this just for a redirect?

classes/Warpcast.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from "zod";
+
+// Neynar's api
+const apiUrl = "https://api.neynar.com";
+const apiKey = process.env.NEYNAR_API_KEY as string;
```
joaquim-verges, line 5: who's key is this? if we're running this in prod we need a proper account - can you ask Mike to set one up for us?

PR #2188, opened 2024-02-08
Fix useIsLoggedInUser

nachoiacovino:


arcoraven (approved):
The boolean changes to useLoggedInUser.ts make sense to me

PR #2438, opened 2024-04-02
chore: revamped billing failed payments system

gerrysaporito:
## PR-Codex overview
This PR introduces changes related to billing sessions, pricing sections, and onboarding flows. It includes updates to billing alerts, pricing components, and onboarding steps.

### Detailed summary
- Added a `billingSession` function in the cache keys
- Updated billing-related links and components
- Adjusted billing pricing and payment methods handling
- Improved error handling for payment verification failures

> The following files were skipped due to too many changes: `src/components/onboarding/index.tsx`, `src/pages/dashboard/settings/billing.tsx`, `src/components/settings/Account/Billing/ManageButton.tsx`, `src/lib/billing.ts`, `src/components/onboarding/PaymentForm.tsx`, `src/components/settings/Account/Billing/alerts/RecurringPaymentFailureAlert.tsx`, `src/components/settings/Account/Billing/alerts/Alert.tsx`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`


jnsdls:
#### 📝 Changed routes:


| [**…board/settings/billing**](https://thirdweb-22i0p37fy.thirdweb-preview.com/dashboard/settings/billing) | [**/**](https://thirdweb-22i0p37fy.thirdweb-preview.com/) |
|:-:|:-:|
|<a href="https://thirdweb-22i0p37fy.thirdweb-preview.com/dashboard/settings/billing">Screenshot of /dashboard/settings/billing</a><br /><sup><a href="https://images.weserv.nl?url=https://sjc.microlink.io/H23xCYl7nrKMbgNhNDMsvgbNpxcXyUck2__YL7Cdr4qETCPu-a1IqGDUIzHkUMtx_myO-j_s5_PcABCY2m39ug.png">(view full size)</a> |<a href="https://thirdweb-22i0p37fy.thirdweb-preview.com/">Screenshot of /</a><br /><sup><a href="https://images.weserv.nl?url=https://sjc.microlink.io/4jvMOX3JcJAIsqXD_vE3-c-4Isxneg_Dv0PK2LZFEiU7HPVTGSdVR_6RioM_j7sZQr0gBHy7D-Aoyq8yFyev3Q.png">(view full size)</a> |







Commit a98fbc041fa50946e84f7faea45ad8c3cfcb536a (https://thirdweb-22i0p37fy.thirdweb-preview.com).

gerrysaporito:
- ✅ Code Complete
- ✅ Testjam
- Code Review - @jnsdls can I get a review on this?

src/components/settings/Account/Billing/index.tsx
```diff
@@ -74,7 +72,7 @@ export const Billing: React.FC<BillingProps> = ({ account }) => {
       {
         plan,
         feedback,
-        useTrial: !!claimedGrowth && validPayment,
+        useTrial: !account?.trialPeriodEndedAt,
```
edwardysun, line 75: Can you quickly explain this logic change?

edwardysun: Feels like it's very different from the prior logic, and there's not a lot of accompanying changes. Curious to know how you arrived at the new logic

src/components/settings/Account/Billing/alerts/ServiceCutoffAlert.tsx
```diff
@@ -0,0 +1,148 @@
...
+};
+
+export const ServiceCutoffAlert: React.FC<ServiceCutoffAlertProps> = ({
+  onDismiss,
+  affectedServices = [],
+  paymentFailureCode = "bank_decline",
+}) => {
+  // TODO: We should find a way to move this deeper into the
+  // TODO: ManageBillingButton component and set an optional field to override
+  const [paymentMethodSaving, setPaymentMethodSaving] = useState(false);
```
edwardysun, line 33: `useState<boolean>`

edwardysun:
`RecurringPaymentFailureAlert` and `ServiceCutoffAlert` look remarkably similar. Can these be generalized?

src/components/settings/Account/Billing/alerts/Alert.tsx
```diff
@@ -0,0 +1,384 @@
...
+    | "serviceCutoff"
+    | "recurringPayment"
+    | "paymentVerification"
+    | "usage";
+};
+
+export const BillingAlerts = () => {
+  const { isLoggedIn } = useLoggedInUser();
+  const usageQuery = useAccountUsage();
+  const meQuery = useAccount({ refetchInterval: 1000 });
```
edwardysun, line 46: This will poll every 1000ms forever. The parameter also takes in a function that allows you to specify a condition to stop polling

jnsdls: yes this one lets please fix

src/components/settings/Account/Billing/alerts/ServiceCutoffAlert.tsx
jnsdls: not required, `false` is inferred to `boolean` correctly so does not require the annotation

src/components/settings/Account/Billing/index.tsx
gerrysaporito: Yeah the orignal if can be broken into 2 parts:
1. claimedGrowth - this used to check to see if the browser has a localhost variable saved that gets set when the user originally claimed the growth package. This logic had a bug where if you logged into a different account on the same browser you wouldn't be given growth free trial. Now we just check if the trial period was claimed directly on the account which is the same thing but account-scoped
2. validPayment - this is validation logic that should happen server-side so it's not necessary to have here (I figured less code client-side the better). I could've done the same thing for the `claimedGrowth` variable (it does happen server side) but I didn't wanna change the behaviour too much from what it is now. The claimedGrowth logic occurs in other places in the billing flow too but validPayment is much rarer so it made sense to remove it from this scope

TL;DR - the old logic had a bug with sharing accounts on the same browser and unecessary business-logic client-side

src/components/settings/Account/Billing/alerts/Alert.tsx
gerrysaporito: Done ✅

src/components/onboarding/ChoosePlan.tsx
```diff
@@ -13,6 +17,8 @@ export const OnboardingChoosePlan: React.FC<OnboardingChoosePlanProps> = ({
 }) => {
   const trackEvent = useTrack();
   const mutation = useUpdateAccount();
+  const { data: account } = useAccount();
+  const canTrialGrowth = !account?.trialPeriodEndedAt;
```
jnsdls, line 21: this does not take into account that `useAccount()` is async and may be loading, meaning while the account is loading `canTrialGrowth` will always be true

If that is the intended behavior that's fine but calling it out because it's not explicitly specified and rather a side-effect of the `!` + optional chaining here

src/components/onboarding/PaymentForm.tsx
```diff
@@ -1,10 +1,19 @@
...
+  Flex,
+  Spinner,
+} from "@chakra-ui/react";
 import {
   useElements,
   useStripe,
   PaymentElement,
 } from "@stripe/react-stripe-js";
+import { PaymentVerificationFailureAlert } from "components/settings/Account/Billing/alerts/PaymentVerificationFailureAlert";
+import message from "components/settings/ApiKeys/message";
```
jnsdls, line 16: is this used?

also this import makes it look like we're using `export default` inside this file we're importing which we should not do except for extraordinary circumstances

src/components/onboarding/index.tsx
```diff
@@ -126,7 +127,7 @@ export const Onboarding: React.FC = () => {
         {
           onSuccess: (data) => {
             if (!skipBilling(data as Account)) {
-              setState(claimedGrowth ? "skipped" : "plan");
+              setState(account?.trialPeriodEndedAt ? "skipped" : "plan");
```
jnsdls, line 130: again calling out `account` can be async, is this the explicit behavior we want here?

src/components/onboarding/index.tsx
```diff
@@ -174,7 +175,7 @@ export const Onboarding: React.FC = () => {
       );
     }
     // skip when going thru claiming trial growth
-    else if (claimedGrowth) {
+    else if (account?.trialPeriodEndedAt) {
```
jnsdls, line 178: account can be async, is that OK here?

src/components/settings/Account/Billing/ManageButton.tsx
```diff
@@ -24,59 +24,70 @@ export const ManageBillingButton: React.FC<ManageBillingButtonProps> = ({
...
+    if (buttonLabel === "manage" && !sessionUrl) {
       mutation.mutate(undefined, {
         onSuccess: (data) => {
           setSessionUrl(data.url);
         },
       });
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [onClick]);
+  }, [buttonLabel]);
```
jnsdls, line 76: this is an extremely weird pattern

I assume the behavior we want here is:

if `buttonLabel` is `manage` and we do not have a `sessionUrl` -> create a session and return the `sessionUrl`

but we've (for some bizarre reason) split this into a `useState` and `useMutation` when what this *should* be is a single `useQuery` that returns a sessionUrl and is `enabled` when buttonLabel === 'manage'

this will absolutely cause bugs down the line, we have to fix this

src/components/settings/Account/Billing/ManageButton.tsx
```diff
@@ -24,59 +24,70 @@ export const ManageBillingButton: React.FC<ManageBillingButtonProps> = ({
...
+      buttonLabel = "addAnotherPayment";
+      buttonText = "Add another payment method →";
+      break;
+    }
+    default: {
+      buttonLabel = "addPayment";
+      buttonText = "Add payment method";
+      break;
+    }
+  }
```
jnsdls, line 53: this... technically works but it would be a lot cleaner if you just wrapped it in either a `useMemo()` or just an iife, you wouldn't have to do the `let` + assign after dance

src/components/settings/Account/Billing/Pricing.tsx
```diff
@@ -1,31 +1,29 @@
...
-import { AccountPlan } from "@3rdweb-sdk/react/hooks/useApi";
+import { Account, AccountPlan } from "@3rdweb-sdk/react/hooks/useApi";
 import { PricingCard } from "components/homepage/sections/PricingCard";
 import { useMemo } from "react";
 import { CONTACT_US_URL } from "utils/pricing";
 import { remainingDays } from "utils/date-utils";
 
 interface BillingPricingProps {
-  plan: AccountPlan;
+  account: Account;
```
jnsdls, line 9: the word `Account` means a specific thing in our `v5` sdk and I don't love that we're overloading this here, better would be `ThirdwebAccount` or something here

src/components/settings/Account/Billing/Pricing.tsx
```diff
@@ -1,31 +1,29 @@
...
   validPayment,
   paymentVerification,
   invalidPayment,
-  trialPeriodEndedAt,
   loading,
-  canTrialGrowth,
   onSelect,
 }) => {
+  const { plan, trialPeriodEndedAt } = account ?? {};
+  const canTrialGrowth = !trialPeriodEndedAt;
```
jnsdls, line 26: looks cary because of the lots of conditionals put together

jnsdls: ... if all we want to know here in the end is `canTrialGrowth` can we just pass that in as a prop to begin with and do this logic above?

src/components/settings/Account/Billing/alerts/Alert.tsx
```diff
@@ -0,0 +1,393 @@
...
+  const meQuery = useAccount({
+    refetchInterval: (account) =>
+      [
+        AccountStatus.InvalidPayment,
+        AccountStatus.InvalidPaymentMethod,
+        AccountStatus.PaymentVerification,
+      ].includes(account?.status as AccountStatus)
+        ? 1000
+        : false,
+  });
```
jnsdls, line 54: nice!

src/lib/billing.ts
```diff
@@ -0,0 +1,95 @@
+enum PaymentFailureCode {
```
jnsdls, line 1: enums are evil, string unions are the way to go (not blocker just commentary)

jnsdls (changes requested):
a couple of things we have to fix before we can ship

gerrysaporito:
> `RecurringPaymentFailureAlert` and `ServiceCutoffAlert` look remarkably similar. Can these be generalized?

Yup, done ✅

src/components/settings/Account/Billing/Pricing.tsx
gerrysaporito: Yup this makes sense, updated ✅

src/components/settings/Account/Billing/Pricing.tsx
gerrysaporito: You're right, we probably shouldn't pass in the whole account object if its not needed (only 1-2 properites), reverted ✅

src/components/settings/Account/Billing/ManageButton.tsx
gerrysaporito: That makes sense to me, never thought of it that way. I updated it to use a `useMemo()`, thanks Jonas! ✅

src/components/onboarding/PaymentForm.tsx
gerrysaporito: Updated, good catch. I didn't realize this was importing a new field for message. Will keep this in mind when reviewing future PRs

src/components/onboarding/index.tsx
gerrysaporito: This should be safe because there's a check at the top to see if account is not null (which occurs when we're loading). If this is an anti-pattern or if there's a better way to do this then I can update it though

src/components/onboarding/index.tsx
gerrysaporito: This one is a good callout, I think account _will be_ null here. I see from above that data is supposed to be account so I'll use `data?.trialPeriodEndedAt` instead and test it to see if this is the expected behaviour

Thanks again jonas for this one, good to know/be aware of 🙏

PizzaShift:
Load accuracy sab
Marcus Jon Pacheco <480-299-4506> 
 
  On Fri, Apr 5, 2024 at 14:14, Gerry ***@***.***> wrote:   



@gerrysaporito commented on this pull request.

In src/components/settings/Account/Billing/Pricing.tsx:
>  import { PricingCard } from "components/homepage/sections/PricingCard";
 import { useMemo } from "react";
 import { CONTACT_US_URL } from "utils/pricing";
 import { remainingDays } from "utils/date-utils";
 
 interface BillingPricingProps {
-  plan: AccountPlan;
+  account: Account;

You're right, we probably shouldn't pass in the whole account object if its not needed (only 1-2 properites), reverted ✅

src/components/onboarding/ChoosePlan.tsx
gerrysaporito: This should be safe because a user who signs up and is seeing this hasn't had a chance to claim the growth plan yet. The default in this case should be true I'd say

Also we have a server-side check to see if they're eligible for the free trial so if a bad actor attempts to abuse this we'll just charge them the full price (no free trial)

Good catch though because this is something that is pretty easy to miss🙏

src/components/settings/Account/Billing/ManageButton.tsx
gerrysaporito: Good feedback, I updated it to do the pattern you suggested. I didn't know how convenient react-query is and it really cleaned up this logic here

Thanks for the idea @jnsdls!

src/components/onboarding/ChoosePlan.tsx
gerrysaporito: Actually I guess in this case, it should always be true

I'll update this to make leave it true and pass it through

gerrysaporito: Done ✅

src/components/onboarding/PaymentForm.tsx
```diff
@@ -41,7 +50,11 @@ export const OnboardingPaymentForm: React.FC<OnboardingPaymentForm> = ({
     const { error: submitError } = await elements.submit();
```
jnsdls, line 50: actually not new code but above this line where we do
```
if (!stripe || !elements) {
      return;
    }
```
    
this means if stripe or elements is not available we never re-set `setSaving` and the user will be stuck - we should move the `setSaving(true)` below the early return statement

jnsdls (changes requested):
found one more thing we should fix, then just waiting for @edwardysun to sign off on his comment

src/components/onboarding/PaymentForm.tsx
gerrysaporito: Great thanks @jnsdls , made the update so its good to go! Will re-ping @edwardysun to sign off and we can merge it in tomorrow

jnsdls (approved):
LGTM from my side at this point

edwardysun (approved):

PR #2460, opened 2024-04-09
Update packages

kumaryash90:
Custom gas settings for create2 deployment on any EVM (https://github.com/thirdweb-dev/js/pull/2678)


jnsdls (changes requested):
underlying `thirdweb` package branch fails tests, has to be fixed first

jnsdls:
outdated as part of #2512

PR #2468, opened 2024-04-10
feat: superchain embedded form (farcaster frame)

easyrun42:
## PR-Codex overview
The focus of this PR is to update the Superchain frame metadata and form handling logic for a smoother user experience.

### Detailed summary
- Updated frame images and metadata in `superchain.tsx`
- Added functions for different frame metadata types in `superchain-embed-frame.ts`
- Implemented form handling logic in `SuperchainFormFrame.ts`
- Created API endpoint for handling frame requests in `frame.ts`

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`


src/classes/SuperchainFormFrame.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  static getEmail = (email: string) => {
+    return z.string().email().parse(email);
+  };
+
+  static getWebsiteUrl = (url: string) => {
+    const regex =
+      // eslint-disable-next-line no-useless-escape
+      /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,})([\/\w \.-]*)*\/?$/;
```
joaquim-verges, line 90: to validate a url you can just do `new URL(theUrl)` in a try catch

src/pages/superchain/frame.tsx
```diff
@@ -0,0 +1,50 @@
...
+            {
+              url: ogImageUrl,
+              width: 1200,
+              height: 630,
+              alt: title,
+            },
+          ],
+        }}
+      />
+      <Head>
```
joaquim-verges, line 37: how come this isnt on the actual superchain landing page? that way when ppl share that it will always render as a frame

src/pages/api/superchain/frame.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    if (type === "redirect") {
+      return redirectResponse(dashboardUrl, 302);
+    }
+
+    return errorResponse("Type not valid", 400);
+  } catch (err) {
+    const errMessage = `Superchain frame embedded form failed`;
+    Sentry.captureException(new Error(errMessage, { cause: err }));
+    return errorResponse("Something went wrong!", 500);
```
joaquim-verges, line 110: what happens if someone enters the wrong chain or an invalid website? would be nice to at least give some feedback on whats wrong no?

src/classes/SuperchainFormFrame.ts
easyrun42: Right. I think we also want to accept `thirdweb.com` without `https://`

src/pages/superchain/frame.tsx
easyrun42: Mhmm good point. We want to save the superchain landing page for later in another frame for now we want to just have it live under `thirdweb.com/superchain/frame`

src/pages/api/superchain/frame.ts
easyrun42: Yeah we don't have designs for those yet. If someone does something wrong it just throws an error, but we hope that no one will type invalid stuff in our frame.

src/classes/SuperchainFormFrame.ts
joaquim-verges: can you just append https:// if not present?

there's a security warning because of that regex

easyrun42: Ah yeah good point

src/pages/api/superchain/frame.ts
easyrun42: Yeah will look at frog and implement it when we do more frames. Is it cool if we merge this one as is ?

src/classes/SuperchainFormFrame.ts
```diff
@@ -85,12 +85,11 @@ export class SuperChainFormFrame {
   };
 
   static getWebsiteUrl = (url: string) => {
-    const regex =
-      // eslint-disable-next-line no-useless-escape
-      /^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,})([\/\w \.-]*)*\/?$/;
-    if (!regex.test(url)) {
-      throw new Error("Invalid URL");
+    if (!url.startsWith("http://") && !url.startsWith("https://")) {
+      return `https://${url}`;
```
joaquim-verges: you're not validating the URL here, should append and then call new URL with the new value

joaquim-verges (approved):

